PR #20, opened 2017-02-01
MIA - Done - Marlon and Stephen

rifenburgh:


Not-A-Sith-Lord:
Should have used random method in javascript to generate new combos on every run.

PR #27, opened 2017-02-10
MIA PT IVAN JORGE CLUE JS GAME

ivan-jorge001:


nickborbe:
2 things
First, not really an issue, I wanted a function that reveals the mystery to the user with a sentence like "it was Dr. Orchid, with the rope, int he billiards room"  Like I said, not a big issue.  All your logic looks good.  

1 thing that is a small issue is this line at the end
`var randkiller = Math.floor(Math.random() * suspects.length);
console.log(suspects[randkiller]);`

This will spit out the whole object like this
`{ firstnames: 'Jack',
  Lastname: 'Mustard',
  Color: 'yellow',
  occupations: ' He is a former football player ',
  age: '19',
  description: 'who tries to get by on his former glory.' }`
which is ugly and not really human readable.  
But, if instead you do this
`var randkiller = Math.floor(Math.random() * suspects.length);
console.log(suspects[randkiller].firstnames + suspects[randkiller].Lastname );`
you get 
`Jack Mustard`

PR #29, opened 2017-02-11
MIA PT Colby Allen - Clue

colbyallenc:


nickborbe:
The only issue: 
`function pickCardRooms() {
  var randomNumber = Math.random();
  randomNumber = randomNumber * clueDo[2].length;
  randomNumber = Math.floor(randomNumber);
  return clueDo[2][randomNumber];`

If instead you do:
  `return clueDo[2][randomNumber].roomName;`
for the last line, then instead of ` { roomName: 'kitchen',
  space: 'medium',
  color: 'orange',
  concurrency: '4' }`
you get 
`kitchen`
Which is prettier (and also part of what we were trying to get you to see.

However, you get major bonus points for the secret envelope you made with the secret confirmation code.

PR #31, opened 2017-02-11
MIA PT Alice Guan - Clue

aliguan:


nickborbe:
Good work.  Looks awesome.  I see that you took the liberty of making the rooms strings instead of objects.  Good idea.

PR #32, opened 2017-02-11
MIA PT Adel Acebal

aacebal:


nickborbe:
Good work.  
I see that you recognized that all we need from the Rooms and Weapons is their name, so you made them strings instead of objects.  Good idea.  
Everything looks good.  Keep it up.

PR #33, opened 2017-02-11
MIA PT Orestes-Garcia

ogm710811:


nickborbe:
Good work.  
The logic looks good.  You defined all the people, rooms and weapons and you picked one of each at random and saved them to your secret envelope.  
The only problem, you got a little fancy with revealing the mystery and forgot to tell the user which room and which weapon, you only told them the name of the killer.  
Other than that, looks good.

PR #34, opened 2017-02-11
MIA PT - Juan Duque

imth3walrus:
Done


nickborbe:
Nice work. 
You made the function that reveals the answer a little over-complicated so now it only tells the user their name but not the room or the weapon.  But I can see you picking a random weapon and random room from the cards and saving them to your secret envelope so all the logic looks good.

PR #35, opened 2017-02-11
done

sholpank:


clue.js
```diff
@@ -0,0 +1,93 @@
...
+  return cards[randomCard];
+}
+
+var myCard = randomCard();
+console.log(myCard);
+
+
+//2.Create another function that calls the method above once for each card stack- working!
+
+function randomCard(cards) {
```
nickborbe: This function doesn't actually pick a random card.  It just says `this is a random card Suspects!`

nickborbe:
You created all the Suspects, Weapons, and Rooms correctly, and you even created a function that picks a random number.  But you need to make that function pick 1 random suspect, 1 random weapon, and 1 random room.

PR #36, opened 2017-02-11
MIA PT Sherwin Garcia

sherwino:
I believe everything works now.


nickborbe:
umm.... yeah so I think you completed the assignment... plus bonus features...

PR #37, opened 2017-02-12
MIA PT CLUE Aleksandra Boskovic

sandrabosk:


nickborbe:
It looks like you got a little stuck on the 'confidential envelope' thing.  I don't care much about that.  All the logic before that part looks good and that's what matters.  I'll give you bonus point for that part where you confirm the user's name.

PR #38, opened 2017-02-12
MIA PT- Jessica Palacios

MzLovelyj:


nickborbe:
Looks good.  You're 95% there.  
It looks like you got stuck on the whole 'confidential envelope' thing.  I'll give you bonus point for var x = confirm("Case File Confidential!");  

The directions were a little unclear, all we wanted was a separate function to pick the 3 cards at random and then another function to reveal the mystery to the user.  All your logic looks good though and that's what matters.

PR #39, opened 2017-02-14
PT MIA Gustavo Gutierrez

ImGuss:
done


nickborbe:
Perfect.  
Good comments as well (not that complicated - I know, but keep up the habit and you'll be glad you did.)

PR #41, opened 2017-02-14
MIA PT WADSON CLUE

iengineer:
I tried.


nickborbe:
Nice work. 
You got everything except a 'pretty' function that reveals the mystery to the user.  But you got all the cards and picking them out at random effectively so that's the important part of the assignment.

PR #48, opened 2017-04-05
BCN-Sergi-Garcia-Phillip-Kickum

nolegendhere:


solution.js
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function showEnvelope(envelopeArg){
+
+  console.log(envelopeArg);
+}
+
+envelope = makeEnvelope(cardStacks);
+
+showEnvelope(envelope);
```
tawebbcn: #checked- well done

PR #49, opened 2017-04-05
BCN "cluedo done" "Alejandro Diaz","Héctor Fernández"

atorau:


starter-code/cluedo.js
```diff
@@ -0,0 +1,69 @@
...
+  weapons.forEach(function(weapon){
+    if(weapon.name == weaponElected){
+      weapon[key]=value;
+    }
+  });
+}
+//weapons[1].status="down";
+
+weaponAtribute(weaponElected,'status','down');
+console.log(weapons);
```
tawebbcn: #checked

PR #51, opened 2017-04-05
BCN-Ryan-Giuseppe

gfrr:


objects.js
```diff
@@ -0,0 +1,81 @@
...
+    return key[randomNumber];
+
+}
+
+function secretEnvelope(suspects, weapons, rooms){
+  return "It was " + randomCard(suspects).name + " in the " + randomCard(rooms) + " with the " + randomCard(weapons);
+
+}
+
+console.log(secretEnvelope(suspects, weapons, rooms));
```
tawebbcn: #checked. Try to name the keys of the objects with a name that represents the value of the key.

PR #52, opened 2017-04-05
BCN-clueGame-Ryan+Guiseppe

ryan-jones:
Finished Clue objects project

### DIFF
--- a/ClueGamePP.js
+++ b/ClueGamePP.js
@@ -1,0 +1,76 @@
+var suspects = {
+    1:{
+      name:"Jacob Green",
+      color: "his color is green.",
+      attribute: "He has a lot of connections and is always willing to help people out -- for a price."},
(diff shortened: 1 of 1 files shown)


ClueGamePP.js
```diff
@@ -0,0 +1,76 @@
...
+    return key[randomNumber] ? key[randomNumber] : randomCard(key);
+
+}
+
+
+function secretEnvelope(suspects, rooms, weapons){
+  return "It was " + randomCard(suspects).name + " in the " + randomCard(rooms) + " with the " + randomCard(weapons);
+}
+
+console.log(secretEnvelope(suspects, rooms, weapons) );
```
tawebbcn: #checked. Try to name the keys of the objects with a name that represents the value of the key.

PR #67, opened 2017-04-05
BCN-Cyril-Pierre

caa1982:


clue_code.js
```diff
@@ -0,0 +1,54 @@
...
+		{ 8: {name:"Pistol", weight:"1kg", color:"blue" , damage:90 } }
+	],
+
+	rooms: ["Dining room", "Conservatory", "Kitchen", "Study", "Library", 
+	"Billiard room", "Lounge", "Ballroom", "Hall", "A spa", "Living room",
+	"Observatory", "Theater", "Guest house", "A patio"],
+		
+}
+
+
```
matilore: The following 3 functions are identical except for the array you are basing the picking on. You should create just a single function, call it 3 times an each time you pass a different array.

PR #103, opened 2017-06-14
BCN: Sara and Rachel

RachelReneeFlores:
We forgot to change all of the suspect data and only changed the first names. Our bad :/

### DIFF
--- a/clue.js
+++ b/clue.js
@@ -1,0 +1,62 @@
+// Sara and Rachel's project
+
+var cards = {
+    suspects: [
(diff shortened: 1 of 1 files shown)


clue.js
```diff
@@ -0,0 +1,62 @@
...
+  };
+}
+
+var mysteryAnswer = returnMystery(cards);
+
+function revealMystery(answer) {
+  console.log(answer.suspects.firstName + " " + answer.suspects.lastName + " did it with the " + answer.weapons.name + " in the " + answer.houseRooms.name + ".");
+}
+
+revealMystery(mysteryAnswer);
```
tawebbcn: good job!

PR #106, opened 2017-06-14
BCN-julius-micaela

jsur:
done

### DIFF
--- a/cluedo.js
+++ b/cluedo.js
@@ -1,0 +1,148 @@
+var packOfCards = {
+  suspects:
+    [
+        {
+          name: 'mrGreen',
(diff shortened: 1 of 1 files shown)


cluedo.js
```diff
@@ -0,0 +1,148 @@
...
+      {'name': 'patio'},
+    ]
+}
+
+var selectRandomIndex = function(array){
+  return Math.floor(Math.random()*array.length);
+};
+
+function selectRandomCard(object) {
+  var keys = Object.keys(object);
```
matilore: Well though but for the next time you can use the 'for in' loop

PR #109, opened 2017-06-14
BCN - Jorge Avila & Laura

lauravallhonrat:
done


matilore:
I think the exercise could be explained better but it would have been nice to have the mistery solved in a different way. For example by accessing for every random object found the value under a property 'name' you could have that value and use it in order to make a sentence that could be the mistery solved, if this is not clear, ask me later, besides this detail everything good

PR #110, opened 2017-06-14
bcn alejandro + petra

PetraSp:


matilore:
You missed the last part, you didn't save the result in  a 'secret envelop' you just console the results you get from the random functions generator.
You should have used every 'name' property value to make a sentence, store the sentence in a variable and then console log that variable, if you have doubts, ask me later

PR #112, opened 2017-06-14
Alex Buchan Cluedo Exercise

alexbuchan:


lab-javascript-clue/cluedo.js
```diff
@@ -0,0 +1,72 @@
...
+
+
+// Method to randomly select one element from a card stack (General)
+function randomCard(cardList) {
+  var min = 0;
+  var max = cardList.length - 1;
+  var randNum = Math.floor(Math.random() * (max - min));
+  var card = cardList[randNum];
+  return card;
+};
```
tawebbcn: `var max = cardList.length - 1`;   should be `var max = cardList.length`
otherwise you are leaving out the last item in the array.
Since you know the min value is gonna be zero, you could refactor as: 
`var randNum = Math.floor(Math.random() * cardList.length);`

lab-javascript-clue/cluedo.js
```diff
@@ -0,0 +1,72 @@
...
+// Function that calls the method above for each card stack
+
+var allList = [suspectList, weaponList, roomsList];
+
+function randomList(list) {
+  var result = [];
+  for (var i = 0; i < list.length; i+=1) {
+    result.push(randomCard(list[i]));
+  }
+
```
tawebbcn: Nice done 👍

lab-javascript-clue/cluedo.js
```diff
@@ -0,0 +1,72 @@
...
+};
+
+// Function that calls the method above for each card stack
+
+var allList = [suspectList, weaponList, roomsList];
+
+function randomList(list) {
+  var result = [];
+  for (var i = 0; i < list.length; i+=1) {
+    result.push(randomCard(list[i]));
```
matilore: you create the object structure in the best possible way, by giving the same name property to all objects you can automatize your code and make it more efficient. The problem you didn't take full advantage of what you build, instead of accessing the name property in line 69 3 times you can access it directly in the line 56 once

matilore:
well done

PR #115, opened 2017-06-14
BCN Brad & Stephanos Day 3

stephanos7:


Brad_Stephanos.js
```diff
@@ -0,0 +1,152 @@
...
+{space: "Ballroom", frequentedBy:[drOrchid, mrPlum]},
+{space: "Hall", frequentedBy: [mrGreen]},
+{space: "Spa", frequentedBy: [mrsScarlet,mrsPeacock,drOrchid]},
+{space: "Living room", frequentedBy: [mrGreen, mrsScarlet]},
+{space: "Observatory", frequentedBy: [mrGreen, mrMustard]},
+{space: "Theater", frequentedBy: [drOrchid, mrsPeacock]},
+{space: "Guest house", frequentedBy:[mrPlum]},
+{space: "Patio", frequentedBy:[mrPlum, mrMustard]}
+   ];
+
```
tawebbcn: For more readability try put each `key: keyvalue,` in a new line

Brad_Stephanos.js
```diff
@@ -0,0 +1,152 @@
...
+  return Math.floor(Math.random()*array.length);
+};
+
+var cardstacks = [suspects, weapons, rooms];
+var confidentialenvelope =[];
+
+for(i = 0; i < cardstacks.length; i++){
+  var pick = randomiser(cardstacks[i]);
+  confidentialenvelope.push(cardstacks[i][pick]);
+}
```
tawebbcn: You are declaring var pick in each loop, would be better to declare it outside & before the loop. `var pick;`
Then inside just assigning the new value `pick = randomiser(cardstacks[i]);`

For loop its totally fine, could you try to achieve the same result using the forEach method ?

Brad_Stephanos.js
```diff
@@ -0,0 +1,152 @@
...
+function giveMeRoom(){
+  var answer = prompt("Where do you think the dreadful murder was committed? (Dining room, Conservatory,Kitchen,Study,Library,Billiard,Lounge,Ballroom,Hall,Spa,Living room,Observatory,Theater,Guest House, Patio):");
+  if(answer === confidentialenvelope[2].space ){
+    alert("Well done, you have guessed correctly!");
+    roomCorrect++;
+  }else{
+    console.log("\n");
+    console.log("no!");
+  }
+}
```
tawebbcn: giveMeName, weapon, room do exactly the same. Try refactor it in just one function.

Also is good to have consistency in your object structure:
Suspects have `firstName`, weapons have `type`, and rooms have `space`
They all have` name:` you are over complicating with 3 different keys.

Brad_Stephanos.js
```diff
@@ -0,0 +1,152 @@
...
+while(nameCorrect===0){
+  giveMeName();
+}
+while(weaponCorrect===0){
+  giveMeWeapon();
+}
+while(roomCorrect===0){
+  giveMeRoom();
+}
+
```
tawebbcn: Nice done ! 👍 
I think would be better to give it just some tries ( like 3 for each object ), otherwise Game Over!

PR #120, opened 2017-06-28
Paris - Lesly and Mathieu

MacSlymo:


cluedo.js
```diff
@@ -0,0 +1,106 @@
...
+  "observatory",
+  "theater",
+  "guest house",
+  "patio"
+];
+
+//we created the function getRand to select randomly, then the function getCard to select randomly among the cards
+
+function getRand(card) {
+  return Math.random() * (card.length - 0) + 0;
```
posva: don't forget the Math.floor() call to round down the value to an index

posva: ... instead of doing it everywhere else

cluedo.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+var player = {} ;
+
+function getCard(player) {
```
posva: no need to add an argument here, instead, create an empty object inside of the function and return it as it's needed

cluedo.js
```diff
@@ -0,0 +1,106 @@
...
+
+//we created the envelope with one case scenario
+
+var caseFileConfidential = {};
+
+caseFileConfidential = getCard(caseFileConfidential);
+
+//we display the answer properly
+
+console.log("The murderer is " + caseFileConfidential.suspectCard.firstname + " " + caseFileConfidential.suspectCard.lastname + ", who killed with the " + caseFileConfidential.weaponCard + " in the " + caseFileConfidential.houseRoomsCard);
```
posva: You could have put this into a function 🙂

PR #126, opened 2017-06-29
William Bonney-Ache

wba0:


script.js
```diff
@@ -0,0 +1,107 @@
+var people={
```
nickborbe: I might make people an array rather than an object.  That way you can loop through it more easily (since order doesn't matter anyway)

script.js
```diff
@@ -0,0 +1,107 @@
...
+    jackM: {
+  firstName: "Jack",
+  lastName: "Mustard",
+  color: "yellow",
+  occupation: "athlete",
+  age: 40},
+
+};
+
+var weapons = {
```
nickborbe: same thing here

script.js
```diff
@@ -0,0 +1,107 @@
...
+
+var pickedArray = function shuffle(deck){
+  var random = Math.floor(Math.random() * deck.length);
+  var selected = deck[random];
+  return selected;
+};
+
+var pickedObject = function randomProperty (obj){
+  var keys = Object.keys(obj);
+  return obj[keys[keys.length * Math.random() <<0]].name;
```
nickborbe: very clever with the <<

nickborbe:
Thanks for submitting

PR #145, opened 2017-08-16
BCN - Philip Wisner, Anca , Jennie

philipwisner:
@ancadelmar
@jenniedalgren


tawebbcn:
Good job but... Why do you repeat some code twice? :+1:

PR #155, opened 2017-09-06
PAR - Jad Challita - Thibaut Davoult

thibautdavoult:


starter-code/clue.js
```diff
@@ -1,104 +1,256 @@
...
+  array.splice(newIndex, 1);
+  return element;
 }
 
 function pick_mistery() {
-
+  var room = random_selector(rooms);
+  var weapon = random_selector(weapons);
+  var character = random_selector(characters);
+  var mistery = {
```
yacinehmito: You don't need the intermediate variables, but you can keep them if you think it makes the code easier to understand. If you don't want them you directly write:

```
var mistery = {
  room: random_selector(rooms),
  weapon: random_selector(weaponss),
  character: random_selector(characters),
}
```

PR #156, opened 2017-09-06
done - Cluedo - Thomas Beguin / David Pinhas 

yithro:


starter-code/clue.js
```diff
@@ -1,103 +1,155 @@
...
-*/
 
-function random_selector(array) {
 
+function pick_mistery() {
+  var mystery = {};
+  mystery.killer = random_selector(characters);
+  mystery.location = random_selector(rooms);
+  mystery.weapon = random_selector(weapons);
+  return mystery;
```
yacinehmito: You can also do something like this if you want:

```
mystery: {
  killer: random_selector(characters),
  location: random_selector(rooms),
  weapon: random_selector(weapons)
}
```

By the way, I liked that you named the properties `killer` and `location` because they have much more meaning than `character` and `room` 😊 .

PR #158, opened 2017-09-06
PAR - Nicolas BARTHEZ & Mejdi ARFAOUI - Iteration1

nicolasBarthez:
Done!

### DIFF
--- a/starter-code/clue.js
+++ b/starter-code/clue.js
@@ -1,104 +1,149 @@
-/*
 
----- Characters ----
 
-  Mr Green
-  first name: Jacob
-  last name:  Green
-  color: green
-  description: He has a lot of connections
-  age: 45
-  image: https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-  occupation: Entrepreneur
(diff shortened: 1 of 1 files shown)


starter-code/clue.js
```diff
@@ -1,104 +1,149 @@
...
+"Lounge",
+"Ballroom",
+"Hall",
+"Spa",
+"Living Room",
+"Observatory",
+"Theater",
+"Guest House",
+"Patio"
+];
```
yacinehmito: It would have made more sense to use objects instead of strings, but this works too.

starter-code/clue.js
```diff
@@ -1,104 +1,149 @@
...
 
-function pick_mistery() {
 
+
+function pick_mistery(envelop) {
+  var weapon = random_selector(stackCardWeapon);
+  var suspect = random_selector(stackCardSuspect);
+  var room = random_selector(stackCardRoom);
+
+  envelop = [weapon,suspect,room];
```
yacinehmito: An object with 3 properties would have been clearer than an array; but again, this works.

PR #161, opened 2017-09-06
PAR: stuart-gardner reshma

stuartogardner:
Our cluedo game. It works, hurrah!

Although one issue we had, was with the function:  

```
function pick_mistery() {
    return caseFile = {suspect: random_selector(suspects), weapon: random_selector(weapons), room: random_selector(rooms)};
}
```
Initially we had:

```
function pick_mistery() {
    var caseFile = {suspect: random_selector(suspects), weapon: random_selector(weapons), room: random_selector(rooms)};
return caseFile;
}
```
But for some reason this didn't work. Not sure why.

### DIFF
--- a/starter-code/clue.js
+++ b/starter-code/clue.js
@@ -1,103 +1,139 @@
-/*
+var mrGreen = {
+  name: "Jacob Green",
+  colour: "green",
+  description: "He has a lot of connections",
+  age: 45,
+  image: "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
+  occupation: "Entrepreneur",
+};
 
----- Characters ----
+var drOrchid = {
+  name: "Doctor Orchid",
+  colour: "white",
+  description: "PhD in plant toxicology. Adopted daughter of Mr. Boddy",
+  age: 26,
+  image: "http://www.radiotimes.com/uploads/images/Original/111967.jpg",
+  occupation: "Scientist",
+};
 
-  Mr Green
-  first name: Jacob
-  last name:  Green
-  color: green
-  description: He has a lot of connections
-  age: 45
-  image: https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-  occupation: Entrepreneur
+var profPlum = {
+  name: "Victor Plum",
+  colour: "purple",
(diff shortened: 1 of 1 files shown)


starter-code/clue.js
```diff
@@ -1,103 +1,139 @@
...
+             {name: "Spa",},
+             {name: "Living Room",},
+             {name: "Observatory",},
+             {name: "Theatre",},
+             {name: "Guesthouse",},
+             {name: "Patio"},
+             ];
+
+var suspects = [];
+suspects.push(mrMustard, mrsPeacock, missScarlet, profPlum, drOrchid, mrGreen);
```
yacinehmito: You can directly do the following:

```
var suspects = [mrMustard, mrsPeacock, missScarlet, profPlum, drOrchid, mrGreen];
```

starter-code/clue.js
```diff
@@ -1,103 +1,139 @@
...
+
+var weapons =[];
+weapons.push(pistol,trophy,bat,axe,poison,dumbell,candlestick,knife,rope);
 
 function random_selector(array) {
+    var randomNumber = Math.floor(Math.random()*array.length);
+    return array[randomNumber].name;
```
yacinehmito: Do you want to get the name or the object itself?

starter-code/clue.js
```diff
@@ -1,103 +1,139 @@
...
 
 }
 
 function pick_mistery() {
-
+    return caseFile = {suspect: random_selector(suspects), weapon: random_selector(weapons), room: random_selector(rooms)};
```
yacinehmito: This works but it doesn't do what you think it does. I suggest that you do instead:

```
return {
  suspect: random_selector(suspects),
  weapon: random_selector(weapons),
  room: random_selector(rooms)
};
```

By the way, I like that you named the property `suspect` instead of `character`. It has a much more precise meaning.

starter-code/clue.js
```diff
@@ -1,103 +1,139 @@
...
 }
 
 function reveal_mistery(mistery) {
-
+    console.log("Mr Boddy was killed by " + caseFile.suspect + " in the " + caseFile.room + " with the " + caseFile.weapon + ".");
```
yacinehmito: Yes!

PR #163, opened 2017-09-06
PAR - Luda Ombline - Cluedo

ombline:


starter-code/clue.js
```diff
@@ -1,104 +1,147 @@
...
+//console.log(random_selector(characters));
 
-function pick_mistery() {
 
+function pick_mistery() {
+var mistery = {};
+mistery.killer = random_selector(characters);
+mistery.room = random_selector(rooms);
+mistery.weapon = random_selector(weapons);
+return mistery;
```
yacinehmito: You can also do something like this if you want:

```
mystery: {
  killer: random_selector(characters),
  location: random_selector(rooms),
  weapon: random_selector(weapons)
}
```

By the way, I liked that you named the property `killer`instead of `character`. It has much more meaning.

PR #185, opened 2017-10-18
BCN - Juan Cuadra & David Alonso

jcuadrad:
@otidlam

### DIFF
--- a/starter-code/clue.js
+++ b/starter-code/clue.js
@@ -1,104 +1,136 @@
-/*
-
----- Characters ----
-
-  Mr Green
-  first name: Jacob
-  last name:  Green
-  color: green
-  description: He has a lot of connections
-  age: 45
-  image: https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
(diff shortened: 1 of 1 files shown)


tawebbcn:
Very good job!
feedback from Isak

```
function createMystery() {
  var caseFileConfidential = [
    random_selector(characters),
    random_selector(weapon),
    random_selector(Rooms),
  ];
  console.log("Mr. Boddy was murdered!");
  return caseFileConfidential;
}
```
// To-->
With this you can use .character or .weapon instead of referring it an element in an array like [0] or [1].. Both works, but it is a bit easier to read the code with the option below: 
```
function pickMistery(characterArray, weaponArray, roomArray) {
  return {
    character: characterArray[randomSelector(characterArray)].nickname,
    weapon: weaponArray[randomSelector(weaponArray)],
    room: roomArray[randomSelector(roomArray)],
  };
}

```

Great if you can clean up the code before submitting and remove all unnecessary console.logs, like this one: console.log("Mr. Boddy was murdered!");

Great job!

PR #188, opened 2017-10-18
BCN: Desiree and Jean

jnothomb:
@dvillool

### DIFF
--- a/starter-code/clue.js
+++ b/starter-code/clue.js
@@ -1,104 +1,196 @@
-/*
 
----- Characters ----
 
-  Mr Green
-  first name: Jacob
-  last name:  Green
-  color: green
-  description: He has a lot of connections
-  age: 45
-  image: https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-  occupation: Entrepreneur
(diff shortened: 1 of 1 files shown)


tawebbcn:
Good job!

feedback: 
mismatch with variable names (roomStack, roomsStack) just use one of them. 
```
/Users/isaklafleur/Google Drive/Isak/Coding/IronHack/module1/lab-javascript-clue/clue.js:167
  roomsStack[Math.floor(Math.random() * roomStack.length)];
            ^

ReferenceError: roomsStack is not defined
```
// improve this to only console.log the name of the character, name of weapon and name of room
```
function pick_mistery() {
  var character = random_selector(characterStack);
  var weapon = random_selector(weaponStack);
  var room = random_selector(roomStack);
  return [character, weapon, room];
}
```

// You should maybe change the data structure of the character array so you can print out the nickname of the character (like "miss Scarlet" and not only her first name)-->
```
function pick_mistery() {
  var character = random_selector(characterStack).name;
  var weapon = random_selector(weaponStack).name;
  var room = random_selector(roomStack);
  return [character, weapon, room];
}
```

PR #190, opened 2017-10-18
BCN-       Eloi and Misha         failed attempt

MythicalMichael:
@eloigil


starter-code/clue.js
```diff
@@ -102,3 +102,186 @@ function reveal_mistery(mistery) {
...
+"Living Room",
+"Observatory",
+"Theater",
+"Guest House",
+"Patio"
+]
+
+
+
+var rope ={Weight: 10}
```
tawebbcn: This objects need a name property too.

PR #191, opened 2017-10-18
BCN - Sara Morales - Jordi Clotet

joordiii:
@saringa

### DIFF
--- a/starter-code/clue.js
+++ b/starter-code/clue.js
@@ -1,103 +1,87 @@
-/*
 
----- Characters ----
+var mrGreen = {
+  firstName: 'Jacob',
+  lastName: 'Green',
+  color: 'green',
+  description: 'He has a lot of connections',
+  age: 45,
+  image: 'https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg',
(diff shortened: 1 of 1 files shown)


starter-code/clue.js
```diff
@@ -1,103 +1,87 @@
...
+  description: 'He is a former football player who tries to get by on his former glory',
+  age: '62',
+  image: 'https://metrouk2.files.wordpress.com/2016/07/colonel-mustard.jpg',
+  occupation: 'Retired Football player'
+};
+
+
+// Deck of cards
+
+var suspects = ['mrGreen','drOrchid','profPlum','missScarlet','mrsPeacock','mrMustard'];
```
tawebbcn: Watch out! You have to create an array with the variables that you declare before, an array of objects. You are making an array of strings.

starter-code/clue.js
```diff
@@ -1,103 +1,87 @@
...
+var weapons = ['rope','Knife','candleStick','dumbbell','poison','axe','bat','trophy','pistol'];
```
tawebbcn: Same thing that with suspects.

PR #192, opened 2017-10-18
BCN - Dominik & Dafne

dafneolca:
@dlasch089


starter-code/clue.js
```diff
@@ -1,92 +1,125 @@
...
+
+var mistery_envelope = pick_mistery(characters, weapons, rooms);
+reveal_mistery(mistery_envelope);
+
+
+// CODE GIVEN BELOW
+
+
 
 function random_selector(array) {
```
tawebbcn: Your original code would "work" if you get rid of these empty functions (lines 124, 128, 132). The problem is that you have the same functions twice and the second time they appea, they are empty. Anyways, there is a better approach to solve this problem, follow my comments.

starter-code/clue.js
```diff
@@ -1,92 +1,125 @@
...
+  "Living Room",
+  "Observatory",
+  "Theater",
+  "Guest House",
+  "Patio"
+]
+
+
+function random_selector(array) {
+  return Math.floor(Math.random()*array.length);
```
tawebbcn: This function is not calle "random_number" but "random selector". It would be expected to return an element of the array, not an index. You can do that this way:
```
function random_selector(array) {
  var index = Math.floor(Math.random()*array.length);
  return array[index];
}
```
I put it in two lines because is easier to read.

starter-code/clue.js
```diff
@@ -1,92 +1,125 @@
...
+}
+
+function pick_mistery(char, weap, rooms) {
```
tawebbcn: You don't need to pass parameters to this function because is accesing global variables (characters, weapons and rooms). That is not the best practice, but it was specified in the function's signature.

starter-code/clue.js
```diff
@@ -1,92 +1,125 @@
...
+  return [
+    char[random_selector (char)],
```
tawebbcn: Now that random_selector returns an element of the array, you can do this as:
```
function pick_mistery() {
  return [
    random_selector(characters),
    random_selector(weapons),
    random_selector(rooms)
 ];
}
```
Anyways, I would prefer to declare the array first and return it after, this way:
```
function pick_mistery() {
  var mistery = [
    random_selector(characters),
    random_selector(weapons),
    random_selector(rooms)
 ];
 return mistery;
}
```

starter-code/clue.js
```diff
@@ -1,92 +1,125 @@
...
+    weap[random_selector (weap)],
+    rooms[random_selector (rooms)]
+ ];
+ }
+
+function reveal_mistery(mistery) {
+  console.log("the character is: ", mistery[0], " with the weapon: ", mistery[1], "in the room: ", mistery[2]);
```
tawebbcn: Here, you are showing all the information for character and weapon (remember they are objects). Better if you only access to the relevant properties, like this:
```
function reveal_mistery(mistery) {
  console.log("the character is: ", mistery[0].firstName, " with the weapon: ", mistery[1].type, " in the room: ", mistery[2]);
}
```

PR #194, opened 2017-10-18
MAD - PP - Eduardo Rodés (solo today)

erodespons:
I didn't totally understand the last points. As I had already prepared a very nice console.log I have kept it as output!


ta-web-mad:
Buen trabajo Eduardo! Sigue así!

PR #201, opened 2017-10-19
MIA - Alfonso Moncada - Ian Izaguirre - Clue Game

ianizaguirre:


dsoraluz:
Great job on getting the game logic down. 

The most important thing to take from this exercise is how we can think of a big problem and break it down into smaller more manageable problems. This is also your first intro to game logic which is something you will be using at the end of module one when you use all of the things you have learned up to that point to create your very own games!

Diving into the game logic, it is important to note how arrays and objects can help us organize the data that we want the user to interact with as well as the data that will drive forward the logic of the game. Learn these structures well and make yourself familiar with the functions they give you.

PR #212, opened 2017-10-25
Paris Charlotte Exercice Cluedo 24 octobre

Charlottebtmn:


ta-web-paris:
Good job Charlotte.
Instead of writing `6` in the following line, `var randomCharacter = clue.characters[parseInt(Math.random()*6+0)].cardName;`, you should have used the length of an array.
Maxence

PR #213, opened 2017-10-25
PAR @amarc27

amarc27:


ta-web-paris:
Good job.
Instead of writing `6` in the following line, `var randomCharacter = clue.characters[parseInt(Math.random()*6+0)].cardName;`, you should have used the length of an array.

PR #245, opened 2018-01-10
PAR Rithy & Jeff cluedo

JeffCampigli:


ta-web-paris:
Good job, your code is nice:) 

But try to follow the tests with Jasmine and ask for some help if you need to understand them!

Maxence

PR #258, opened 2018-01-10
BCN - Danielle & Rodrigo

danielleleb:
@rodrigodaoud

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,160 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-occupation:   Entrepreneur
 
-drOrchid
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,160 @@
...
+var livingRoom = {name: "Living Room"};
+var observatory = {name: "Observatory"};
+var theater = {name: "Theater"};
+var guestHouse = {name: "Guest House"};
+var patio = {name: "Patio"};
+
 
 // Characters Collection
 var charactersArray = [];
+charactersArray.push(mrGreen)
```
tawebbcn: You can do all the pushes in one line:
```
```

starter-code/src/clue.js
```diff
@@ -1,94 +1,160 @@
...
+
+
+
+function randomSelector(array) {
+    var item = array[Math.floor(Math.random() * array.length)];
+    return item;
+};
+
+
+var killer = randomSelector(charactersArray);
```
tawebbcn: Better do this three selections inside the pickMistery function, it is the only one wich is going to use them.

PR #267, opened 2018-01-10
BCN - Victor and Oscar

victorpellerin:
@oscarjorgejorge

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,131 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,131 @@
...
-// Weapons
+var mrGreen = {
+first_name:   "Jacob",
+last_name:    "Green",
+color:        "green",
+description:  "He has a lot of connections",
+age:          45,
+image:        "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
+occupation:   "Entrepreneur",
+}
```
tawebbcn: Be careful with the indentation and the semicolons!

PR #277, opened 2018-01-24
[MIA] Mosi Rivera

mosileko:
Wassup TA's?

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -84,11 +84,126 @@ name: Guest House
 name: Patio
 */
 
-// Characters Collection
-var charactersArray = [];
 
-// Rooms' Collection
-var roomsArray = [];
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -84,11 +84,126 @@ name: Guest House
 name: Patio
 */
 
+
+
 // Characters Collection
-var charactersArray = [];
+var charactersArray = [
+  {
+    mrGreen:
```
sandrabosk: proper syntax for objects is:
```
var charactersArray = [
  var mrGreen = { ... }
]
```
and so on.. When you change this, you'll have working code.

sandrabosk:
Beside that "small" error that missed your eye, everything else is working properly. Thank you for submitting.

PR #285, opened 2018-01-24
[Mia] [Michael - Elso]

ElsoMichael:
Issues:
- calling on the function in Mystery Envelope to push to the arry.


sandrabosk:
Good work so far, please submit the rest of the code. Thank you.

PR #311, opened 2018-03-21
PAR Bertrand & JB

wavilo:


starter-code/src/clue.js
```diff
@@ -1,94 +1,126 @@
...
 
 // Rooms' Collection
-var roomsArray = [];
+var roomsArray = [dinningRoom, conservatory, kitchen, study, library, billiardRoom, lounge, ballroom, hall, spa, livingRoom, observatory, theater, guestHouse, patio];
 
 // Weapons Collection
-var weaponsArray = [];
+var weaponsArray = [rope, knife, candlestick, dumbbell, poison, axe, bat, trophy, pistol];
+
+function randomSelector(array) {
```
ta-web-paris: You don't need to add "+1" as you are in an array, and the index only go from 0 to 9, the right code should be 👍 
  var nb = Math.floor(Math.random() * array.length);
  return array[nb];
}


Because if you adding one it will sometimes check the index [10] and as there is nothing on index 10 it will return undefined

starter-code/src/clue.js
```diff
@@ -1,94 +1,126 @@
...
+    var nb = Math.floor(Math.random() * (array.length + 1));
+    return array[nb];
+}
+
+function pickMistery(){
+    var misteryEnvelope = [
+    randomSelector(charactersArray),
+    randomSelector(roomsArray),
+    randomSelector(weaponsArray)
+    ]
```
ta-web-paris: Try this piece of code at the end, it will fit with what Jasmine is expecting : 

function pickMistery() {
  var weapon = randomSelector(weaponsArray);
  var room = randomSelector(roomsArray);

  var resultArray = [];
  resultArray.push(killer, weapon, room);
  return resultArray;
}

function revealMistery(mistery) {
  return mistery[0].first_name + ' ' + mistery[0].last_name + ' killed Mr.Boddy using the ' + mistery[1].name + ' in the ' + mistery[2].name + '!!!!';
}

var misteryEnvelope = pickMistery();
console.log(revealMistery(misteryEnvelope));

ta-web-paris:
If you have any questions regarding the end of the exercice feel free to ask ! 

Otherwise good job :) 🥇

PR #313, opened 2018-03-21
PAR Luke & Billy

Yaats:
randomSelector (arrayLength?) - sometimes returning undefined

revealMystery function don't work

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,139 @@
-// Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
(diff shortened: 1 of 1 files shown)


ta-web-paris:
For the array.length isn't a big deal.
It's just Jasmine test.
For pickMistery you tried a lot of things. Which is good.

But the best way to solve the issue is the following:
```js
function pickMistery() {
  var weapon = randomSelector(weaponsArray);
  var room = randomSelector(roomsArray);

  var resultArray = [];
  resultArray.push(killer, weapon, room);
  return resultArray;
}
```
First of all you define your var and **after** you push.
This is the neated way.

Hope this helps!

PR #325, opened 2018-03-21
bcn-fredrik

fredrik62:
did it together with dauf, george and fabian

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,160 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,160 @@
...
+var weaponsArray = [rope,knife,candlestick,dumbbell,poison,axe,bat,trophy,pistol];
+var overallArray =[charactersArray,roomsArray,weaponsArray]
+
+
+// RadomSelector
+
+function randomSelector (Array){
+ var selector = Math.floor(Math.random(0,Array.length));
+ return Array[selector];
+}
```
tawebbcn: Avoid using Array with capital A. Variables should always be defined in camelCase

PR #326, opened 2018-03-21
BCN Fabian

fscheugenpflug:
George, Fredrik, Dauve

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,156 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,156 @@
...
+var weaponsArray = [rope,knife,candlestick,dumbbell,poison,axe,bat,trophy,pistol];
+
+var overallArray =[charactersArray,roomsArray,weaponsArray]
+
+
+// RadomSelector
+
+function randomSelector (Array){
+  var selector = Math.floor(Math.random() * Array.length + 1);
+  return Array[selector];
```
tawebbcn: Do not use the Array word and do not start with capital letters the parameters

PR #329, opened 2018-03-21
BCN - Lennie Zhu

lenniez:
Completed, but with a slightly different methodology than Jasmine tests. mysteryEnvelope has 4 elements instead of three - killer name is split into first and last name, which allows me to access those values in the misteryReveal function

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,224 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,224 @@
...
+  return array[random];
+};
+
+// Mystery
+
+var mysteryEnvelope = [];
+
+pickMistery = function (killer, weapon, room) {
+  var randomKillerFirst = randomSelector(killer).first_name;
+  var randomKillerLast = randomSelector(killer).last_name;
```
tawebbcn: This will cause the first name to not match the second name. The idea is to get one element from the array and then read both values

PR #332, opened 2018-03-21
BCN Ioana

IoanaOrca:
I lost a lot of time with "TypeError: Cannot read property 'length' of undefined" on line 322 from clue.js and had to change the parameters from pickMistery to the real arrays (bad, I know) so I could skip the errors. 

Everything else great.

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -92,3 +92,252 @@ var roomsArray = [];
 
 // Weapons Collection
 var weaponsArray = [];
+
+var mrGreen = {
+    first_name:   "Jacob",
+    last_name:    "Green",
+    color:        "green",
+    description:  "He has a lot of connections",
+    age:          45,
+    image:        "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
+    occupation:   "Entrepreneur"
+  };
+ 
+charactersArray.push(mrGreen);
+
+var drOrchid = {
+    frst_name:    "Doctor,",
+    last_name:    "Orchid,",
+    color:        "white",
+    description:  "PhD in plant toxicology. Adopted daughter of Mr. Boddy",
+    age:          26,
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -92,3 +92,252 @@ var roomsArray = [];
...
+    first_name:   "Jacob",
+    last_name:    "Green",
+    color:        "green",
+    description:  "He has a lot of connections",
+    age:          45,
+    image:        "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
+    occupation:   "Entrepreneur"
+  };
+ 
+charactersArray.push(mrGreen);
```
tawebbcn: You just need it once for each array:
```
charactersArray.push(mrGreen, drOrchid, profPlum, ...);
```

starter-code/src/clue.js
```diff
@@ -92,3 +92,252 @@ var roomsArray = [];
...
+
+  var room15 = {
+        name: "Patio"
+  }
+
+  roomsArray.push(room15);
+
+function randomSelector(aArray){
+    if (aArray.length!==0){
+    return aArray[Math.floor(Math.random()*aArray.length)];
```
```
return aArray[Math.floor(Math.random()*(aArray.length + 1))];
```

starter-code/src/clue.js
```diff
@@ -92,3 +92,252 @@ var roomsArray = [];
...
+   misteryArray[2]=randomSelector(roomsArray);
+   return misteryArray
+}
+
+var misteryEnvelope = pickMistery(charactersArray, roomsArray, weaponsArray)
+
+console.log(misteryEnvelope);
+
+function revealMistery(envelope){
+    return envelope[0].first_name +" "+ envelope[0].last_name+" killed Mr.Boddy using the " + envelope[1].name+" in the " + envelope[2].name +"!!!!";
```
tawebbcn: :+1:

starter-code/src/clue.js

PR #334, opened 2018-03-21
BCN - Sebastian Nugnez

Sebachowa:
I finish the exercise plus 1 bonus.

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,156 @@
+//--------------------------Iteration 1--------------------------
+
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-occupation:   Entrepreneur
 
-drOrchid
-frst_name:   Doctor
-lst_name:    Orchid
-color:        white
-description:  PhD in plant toxicology. Adopted daughter of Mr. Boddy
-age:          26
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,156 @@
...
+var roomsArray = [];
+roomsArray.push(dinningRoom, conservatory, kitchen, study, library, billboardRoom, lounge, ballroom, hall, spa, livingRoom, observatory, theater, guestHouse, patio)
+
+//--------------------------Iteration 2--------------------------
+
+//Create a method randomSelector to randomly select one element from a card stack. 
+//The method should receive an array as an argument, and return randomly one of the elements of the array.
+
+function randomSelector(array) {
+  return array[Math.floor(Math.random() * array.length)];
```
tawebbcn: ```return array[Math.floor(Math.random() * (array.length + 1))];```

starter-code/src/clue.js
```diff
@@ -1,94 +1,156 @@
...
+//We need to create a pickMistery method that will call randomSelector for each card stack,
+//and return an array with the 3 picked cards, a character, a weapon and a room. 
+//Our mystery should be stored on a misteryEnvelope variable.
+
+function pickMistery() {
+  var characterFound = randomSelector(charactersArray);
+  var weaponFound = randomSelector(weaponsArray);
+  var roomFound = randomSelector(roomsArray);
+
+  var misteryEnvelope = [characterFound, weaponFound, roomFound];
```
tawebbcn: misteryEnvelope should be declared outside

starter-code/src/clue.js

PR #335, opened 2018-03-21
BCN-Miriam-Aparicio

MiriamAparicio:


starter-code/src/clue.js
```diff
@@ -1,94 +1,139 @@
...
-mrMustard
-first_name:   Jack
-last_name:    Mustard
-color:        yellow
-description:  He is a former football player who tries to get by on his former glory
-age:          62
-image:        https://metrouk2.files.wordpress.com/2016/07/colonel-mustard.jpg
-occupation:   Retired Football player
+
+function Character (first_name,last_name, color, description, age, image, occupation){
```
tawebbcn: :metal:

starter-code/src/clue.js
```diff
@@ -1,94 +1,139 @@
...
+weaponsArray.push(rope,knife,candlestick,dumbbell,poison,axe,bat,trophy,pistol);
+
+//method to get a Random selector
+function randomSelector (array) {
+
+  var selector;
+  var minNum = 0;
+  var maxNum = array.length;
+  if (array.length > 0) {
+    selector = Math.floor(Math.random()*(array.length - minNum))+ minNum;
```
tawebbcn: selector is not declared

tawebbcn:
```
selector = Math.floor(Math.random()*(array.length + 1));
```

starter-code/src/clue.js
```diff
@@ -1,94 +1,139 @@
...
+    return array[selector];
+  }
+
+}
+
+//create a mystery method and stored it on a mistery envelope
+
+function pickMistery () {
+
+  var misteryEnvelope = [];
```
tawebbcn: ```var misteryEnvelope = [];``` should be declared outside the function

starter-code/src/clue.js
```diff
@@ -1,94 +1,139 @@
...
+  var character = randomSelector(charactersArray);
+  var weapon = randomSelector(weaponsArray);
+  var room = randomSelector(roomsArray);
+  misteryEnvelope.push(character, weapon, room);
+  return misteryEnvelope;
```
tawebbcn: You don't need to return misteryEnvelope because you are already pushing it in the function.

starter-code/src/clue.js
```diff
@@ -1,94 +1,139 @@
...
+
+  return killer + ' killed Mr.Boddy using the ' + weapon + ' in the ' + place + '!!!!';
+  
+}
+
+/*   *BONUS JS*
+- create a function that filter an array by any field
+- create a function that show the characters in alphabetic order*/
+
+function filterByAnyField (array, value, field) {
```
tawebbcn: :+1:

starter-code/src/clue.js

PR #336, opened 2018-03-22
BCN - Mihai Costantea

cosmihai:


starter-code/src/clue.js
```diff
@@ -84,11 +84,163 @@ name: Guest House
...
+  {name: 'Patio'},
+];
+
+//------misteryEnvelope------
+
+var misteryEnvelope = [];
+
+//-----Random selector
+function randomSelector(array) {
+  var index = Math.floor(Math.random()*array.length);
```
```
var index = Math.floor(Math.random() * (array.length + 1));
```

starter-code/src/clue.js
```diff
@@ -84,11 +84,163 @@ name: Guest House
...
+}
+
+//--------pickMistery------
+
+function pickMistery(card1, card2, card3) {
+  
+  misteryEnvelope.push(randomSelector(card1));
+  misteryEnvelope.push(randomSelector(card2));
+  misteryEnvelope.push(randomSelector(card3));
+  return misteryEnvelope;
```
tawebbcn: You don't need to return misteryEnvelope because you are already pushing it in the function.

starter-code/src/clue.js

PR #338, opened 2018-03-22
BCN - RAPHA MONTENEGRO

raphamontenegro:


starter-code/src/clue.js
```diff
@@ -84,11 +84,188 @@ name: Guest House
...
+  dumbbell,
+  poison,
+  axe,
+  bat,
+  trophy,
+  pistol
+];
+
+function randomSelector (array) {
+  var randomNumber = Math.floor((Math.random() * array.length) );
```
```
var random = Math.floor(Math.random() * (array.length + 1));
```

starter-code/src/clue.js
```diff
@@ -84,11 +84,188 @@ name: Guest House
...
+};
+
+var misteryEnvelope = [];
+
+function pickMistery () {
+  var randomCharacter = randomSelector(charactersArray);
+  var randomWeapon = randomSelector(weaponsArray);
+  var randomRoom = randomSelector(roomsArray);
+  misteryEnvelope = [randomCharacter, randomWeapon, randomRoom] 
+  return misteryEnvelope
```
tawebbcn: If you are changing the value of mysteryEnvelope inside the function, you don't need to return anything.

```
var misteryEnvelope;

function pickMistery () {
    var randomCharacter = randomSelector(charactersArray);
    var randomWeapon = randomSelector(weaponsArray);
    var randomRoom = randomSelector(roomsArray);
    misteryEnvelope = [randomCharacter, randomWeapon, randomRoom];
}

```

If you want to use return you could do: 

```
function pickMistery () {
    var randomCharacter = randomSelector(charactersArray);
    var randomWeapon = randomSelector(weaponsArray);
    var randomRoom = randomSelector(roomsArray);
    var array = [randomCharacter, randomWeapon, randomRoom];
    return array;
}
var misteryEnvelope = pickMistery ();
```

starter-code/src/clue.js

PR #339, opened 2018-03-22
BCN - Matias Pierretti

Matuki92:


starter-code/src/clue.js
```diff
@@ -92,3 +92,186 @@ var roomsArray = [];
...
+
+weaponsArray.push(rope,knife,candlestick,dumbbell,poison,axe,bat,trophy,pistol);
+
+roomsArray.push(dinningRoom,conservatory,kitchen,study,library,billiardRoom,lounge,ballRoom,hall,spa,livingRoom,observatory,theater,guestHouse,patio);
+
+//Functions
+
+function randomSelector(array){
+
+  var random = Math.floor(Math.random()*array.length);
```
```
var random = Math.floor(Math.random() * (array.length + 1));
```

starter-code/src/clue.js
```diff
@@ -92,3 +92,186 @@ var roomsArray = [];
...
+
+  return array[random];
+
+}
+
+function pickMistery(){
+
```
tawebbcn: this empty lines are weird

starter-code/src/clue.js

PR #340, opened 2018-03-22
BCN - JordiCalzada

jordicalzada:
I did not understand de exercice.


starter-code/src/clue.js
```diff
@@ -85,10 +85,128 @@ name: Patio
...
+last_name:    'Scarlet',
+color:        'red',
+description:  'She is an A-list movie star with a dark past',
+age:          31,
+image:        'https://metrouk2.files.wordpress.com/2016/07/miss-scarlett.jpg',
+occupation:   'Actor'
+}
+
+var mrsPeacock = {
+
```
tawebbcn: weird empty line

starter-code/src/clue.js
```diff
@@ -85,10 +85,128 @@ name: Patio
...
+name: 'Theater',
+name: 'Guest House',
+name: 'Patio'
+};
+
 var weaponsArray = [];
+
+
+function randomSelector(array){
+    return array[Math.floor(Math.random()*Array.length)];
```
```
return array[Math.floor(Math.random()*(Array.length + 1))];
```

starter-code/src/clue.js
```diff
@@ -85,10 +85,128 @@ name: Patio
...
+}
+randomSelector(mrGreen);
```
tawebbcn: ```randomSelector(charactersArray)```

starter-code/src/clue.js
```diff
@@ -85,10 +85,128 @@ name: Patio
...
+var misteryEnvelope = [];
+function pickMistery(array){
+    
+    array.forEach(function randomSelector(array) {
+        misteryEnvelope = array;
+        return array;
+        
+    });
+}
+pickMistery(mrGreen);
```
tawebbcn: you should be passing the three arrays as arguments and then returning an array with a character, a weapon and a room

starter-code/src/clue.js

PR #341, opened 2018-03-22
BCN - George

GeorgeCoupland:
suppp


starter-code/src/clue.js
```diff
@@ -1,88 +1,87 @@
 // Characters
-/*
+
 mrGreen
```
tawebbcn: ```var mrGreen = {```

starter-code/src/clue.js
```diff
@@ -1,88 +1,87 @@
...
-name: rope        ---   weight: 10
-name: knife       ---   weight: 8
-name: candlestick ---   weight: 2
-name: dumbbell    ---   weight: 30
-name: poison      ---   weight: 2
-name: axe         ---   weight: 15
-name: bat         ---   weight: 13
-name: trophy      ---   weight: 25
-name: pistol      ---   weight: 20
+name: 'rope'-- - weight: 10
```
```
var rope = {
    name: 'rope'
}
```

starter-code/src/clue.js
```diff
@@ -92,3 +91,6 @@ var roomsArray = [];
 
 // Weapons Collection
 var weaponsArray = [];
+
+
+var mysteryCharacters = Math.floor(Math.random * charactersArray.length);
```
```
Math.floor(Math.random * (charactersArray.length + 1))
```
but charactersArray should be pushed with all characters objects.

PR #372, opened 2018-04-05
FT-Miami Brian Jaimes-Clue

B-riJ:


starter-code/tests/ClueSpec.js
```diff
@@ -42,13 +50,38 @@ describe('Pick a random mistery - pickMistery', function () {
     expect(typeof pickMistery).toBe('function');
   });
 
+  function pickMistery(){
+  var RandomCharacter = randomSelector(charactersArray);
+  var RandomRoom = randomSelector(roomsArray);
+  var RandomWeapon = randomSelector(weaponsArray);
+
+  var mistery = [randomCharacter, randomWeapon, randomRoom];
```
sandrabosk: Let's not make habit of skipping the step where we push objects into arrays. You can do your way since you have 3 objects, but what if you would have 30 or 300 objects.
```
var mistery = [];
mistery.push(randomCharacter, randomWeapon, randomRoom);
```

sandrabosk:
Good job overall BUT we suppose to keep our work in clue.js file, spec file is for testing purposes only and we shouldn't be changing it. Thank you for submitting.

PR #374, opened 2018-04-08
MIA-PTWD-MariahDunn-Clue Objects & Arrays

teepott99:
Not passing all the tests but not sure why.  

Pretty sure var use is out of hand. 

Comments/guidance on this would be much appreciated.


starter-code/src/clue.js
```diff
@@ -1,94 +1,119 @@
...
+    {name: "Theater"},
+    {name: "Guest House"},
+    {name: "Patio"}
+];
+
+function randomSelector(arr){
+    if (arr === []) {
+    return;
+    } else {
+        var random = arr[Math.floor(Math.random() * arr.length)];
```
ta-web-mia: Although this is correct, its best to call the variable above the conditional statement. so that way your else statement looks like else {return random}. Aside from that great job!

PR #426, opened 2018-05-23
BER Cameron & Nesli 

cameronskene:
Did not pass all Jasmine tests, but we think it works.

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,114 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-occupation:   Entrepreneur
-
-drOrchid
-frst_name:   Doctor
-lst_name:    Orchid
-color:        white
-description:  PhD in plant toxicology. Adopted daughter of Mr. Boddy
-age:          26
-image:        http://www.radiotimes.com/uploads/images/Original/111967.jpg
-ocupation:   Scientist
-
-profPlum
-first_name:   Victor
-last_name:    Plum
-color:        purple
-description:  Billionare video game designer
-age:          22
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,115 @@
...
+  "Hall",
+  "Spa",
+  "Living Room",
+  "Observatory",
+  "Theater",
+  "Guest House",
+  "Patio",
+  ];
+  
+  function randomSelector(array) {
```
ta-web-ber: max should be array.length, not array.length -1  :)

```
   
    var min = 0;
    var max = array.length; //array.length, not length -1
    var random = Math.floor(Math.random() * (max - min));
    return array[random];
  }

File: starter-code/src/clue.js [hunk 2]
```diff
@@ -1,94 +1,115 @@
...
-var roomsArray = [];
-
-// Weapons Collection
-var weaponsArray = [];
+  function pickMistery() {
+    misteryEnvelope.push(randomSelector(charactersArray), randomSelector(weaponsArray), randomSelector(roomsArray));
+    return misteryEnvelope;
+  }
+  
+  function revealMistery(misteryEnvelope) {
```
ta-web-ber: Test are very picky although the code is correct. Jasmine was expecting an array of objects, but it got an array of arrays.
You coded:
```
var weaponsArray = [
  ["rope", 10],
  ["knife", 8],
  ["candlestick", 2],
  ["dumbbell", 30],
  ["poison", 2],
  ["axe", 15],
  ["bat", 13],
  ["trophy", 25],
  ["pistol", 20],
  ];

```
but was expecting:

```
var weaponsArray = [
  { name: "rope", weight: 10 },
  { name: "knife", weight: 8 },
  { name: "candlestick", weight: 2 },
  { name: "dumbbell", weight: 30 },
  { name: "poison", weight: 2 },
  { name: "axe", weight: 15 },
  { name: "bat", weight: 13 },
  { name: "trophy", weight: 25 },
  { name: "pistol", weight: 20 }
];

PR #432, opened 2018-05-23
BER-Ayumi/Jens-javascipt-clue

je0705:
Except of one last failure it's working.

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,176 @@
-// Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,176 @@
...
-// Rooms' Collection
-var roomsArray = [];
+    var firstName = misteryEnvelope[0].firstName;
+    var lastName = misteryEnvelope[0].lastName;
+    var weapon = misteryEnvelope[1].name;
+    var place = misteryEnvelope[2].name;
 
-// Weapons Collection
-var weaponsArray = [];
+   return firstname + " " + lastName + "killed Mr.Boddy using the " + weapon + " in the " + place + " !!!"
```
ta-web-ber: should be firstName instead of firstname.
```
return "" + firstName + " " + lastName + " killed Mr.Boddy using the " + weapon + " in the " + place + "!!!"
```
Don't worry about the last error. It works in repl.it and the code is solid :)

PR #437, opened 2018-05-29
PAR - Clement Sirvente&Adrien Belhomme

csirvente:


M-Gadd:
Good job. 

but here its better if you create the array and then push the values, instead of adding them manually 
var charactersArray = [mrGreen,drOrchid,profPlum,missScarlet,mrsPeacock,mrMustard];

PR #442, opened 2018-05-30
PAR nadiah

nadiahachemi:


starter-code/src/clue.js
```diff
@@ -85,10 +54,39 @@ name: Patio
...
+    {name: "candlestick", weight: 2},
+    {name: "dumbbell", weight: 30},
+    {name: "poison", weight: 2},
+    {name: "axe", weight: 15},
+    {name: "bat", weight: 13},
+    {name: "trophy", weight: 25},
+    {name: "pistol", weight: 20},
+];
+
+function getTheMystery(numberOfSuspect, numberOfRooms, numberOfWeapons){
```
ta-web-paris: Instead of putting all in one method, and to stick to the jasmine test requirements you should build different function (and remember that is better to one function for one action), so here is what Marie did, and it's actually a really nice way : 

```js 

cardStacks = [];
cardStacks.push(charactersArray, weaponsArray, roomsArray);

//This function randomly selects a card from a card stack
function randomSelector(cardStack){
  var randomCard = cardStack[Math.floor(Math.random()*cardStack.length)];
  return randomCard;
}

function pickMistery(){
  var mysteryEnvelope = [];
  cardStacks.forEach(function(cardStack){
    var mysteryCard=randomSelector(cardStack);
    mysteryEnvelope.push(mysteryCard);
  })
  return mysteryEnvelope;
};

function revealMistery (mysteryEnvelope){
  return mysteryEnvelope[0].first_name + " "+ mysteryEnvelope[0].last_name + " killed Mr.Boddy using the " + mysteryEnvelope[1].name + " in the " + mysteryEnvelope[2].name + "!!!!";
}

var mysteryEnvelope = pickMistery();
revealMistery(mysteryEnvelope);

```

ta-web-paris:
Good job, as I told you don't forget to run the SpecRunner.html, this is where all the jasmine test appear, and it's a requirement for the exercise (even if your code is working ;) )

If you have any question regarding what we've seen so far, let me know!

PR #444, opened 2018-05-30
PAR Shanshan 

moonchie:
I don't know how to push the killer, room and weapon by the index to the array.
I also don't understand other two errors.

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,181 @@
-// Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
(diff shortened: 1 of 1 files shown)


ta-web-paris:
Some of the errors are just about the positioning of the element in the array so it's not a big deal, but there is one about your array (expect 12 to equal 3) but I don't know where it come from, everything seems ok.. I'll take a look at it.. 

Otherwise your code is good :)

starter-code/src/clue.js
```diff
@@ -1,94 +1,181 @@
...
+    weight: 25,
+};
+var pistol = {
+    name: "pistol",
+    weight: 20,
+};
 
 // Characters Collection
 var charactersArray = [];
+charactersArray.push(mrGreen);
```
ta-web-paris: You can push many element in one line like this : 

```js

charactersArray.push(mrGreen, drOrchid, profPlum, missScarlet, mrMustard, mrsPeacock)

```

Same for the other array :)

PR #445, opened 2018-05-30
PAR Jihad

eloji9:
Again, wasn't able to fully complete the exercise. I understand the reasoning behind it but have real difficulties translating them to the javascript analytics.

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -85,10 +85,90 @@ name: Patio
 */
 
 // Characters Collection
-var charactersArray = [];
+var charactersArray = [
+    "mrGreen",
+    "drOrchid",
+    "profPlum",
+    "missScarlet",
+    "mrsPeacock",
+    "mrMustard",
+];
 
 // Rooms' Collection
-var roomsArray = [];
+var roomsArray = [
+    "Dinning Room",
+    "Conservatory",
+    "Kitchen",
+    "Study",
+    "Library",
+    "Billiard Room",
+    "Lounge",
+    "Ballroom",
+    "Hall",
+    "Spa",
+    "Living Room",
+    "Observatory",
+    "Theater",
+    "Guest House",
+    "Patio",
+];
 
 // Weapons Collection
-var weaponsArray = [];
+var weaponsArray = [
+    "rope",
+    "knife",
+    "candlestick",
+    "dumbbell",
+    "poison",
+    "axe",
+    "bat",
+    "trophy",
+    "pistol",
+];
+
+// declare characters' variables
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -85,10 +85,90 @@ name: Patio
...
+var candlestick = {name: "candlestick", weight:"2"};
+var dumbbell = {name: "dumbell", weight:"30"};
+var poison = {name : "poisons",weight:"2"};
+var axe = {name : "axe", weight:"15"};
+var bat = {name : "bat", weight:"13"};
+var trophy = {name : "trophy" , weight:"25"};
+var pistol = {name : "pistol",weight:"20"};
+
+// random select
+function randomSelector(charactersArray){
```
ta-web-paris: Instead of rewriting the same function like this, keep you code DRY ;) 
You can use it like this  : 

```js 

cardStacks = [];
cardStacks.push(charactersArray, weaponsArray, roomsArray);

//This function randomly selects a card from a card stack
function randomSelector(cardStack){
  var randomCard = cardStack[Math.floor(Math.random()*cardStack.length)];
  return randomCard;
}

```

Basically you are creating an array, that will contain all the other arrays (an array of array :) ) and then you apply a random method to each of the element (forEach method) it will take randomly an element from all the array so you will be able to create your mystery

starter-code/src/clue.js
```diff
@@ -85,10 +85,90 @@ name: Patio
...
+
+console.log(randomSelector(roomsArray));
+
+function randomSelector(weaponsArray){
+    return weaponsArray[Math.floor(Math.random()*weaponsArray.length)];
+}
+
+console.log(randomSelector(weaponsArray));
+
+function pickMistery(){
```
ta-web-paris: To finish the exercise you have to reuse all the method you've create before and combine them so you can create a "new game" : 

```js

function pickMistery(){
  var mysteryEnvelope = [];
  cardStacks.forEach(function(cardStack){
    var mysteryCard=randomSelector(cardStack);
    mysteryEnvelope.push(mysteryCard);
  })
  return mysteryEnvelope;
};

function revealMistery (mysteryEnvelope){
  return mysteryEnvelope[0].first_name + " "+ mysteryEnvelope[0].last_name + " killed Mr.Boddy using the " + mysteryEnvelope[1].name + " in the " + mysteryEnvelope[2].name + "!!!!";
}

var mysteryEnvelope = pickMistery();
revealMistery(mysteryEnvelope);

```

ta-web-paris:
It's okay, it will come by practicing (and you will practice a lot trust me ;) )
Actually, you have a good way to approach the problem, but it looks like you are struggling to put all the piece together, if there is some concept that you don't understand, let me know! 

keep up the good work!

PR #447, opened 2018-05-30
Update clue.js

acmenow:


starter-code/src/clue.js
```diff
@@ -1,94 +1,192 @@
...
+
+ //console.log(pickMistery(charactersArray,weaponsArray,roomsArray));
+
+ //var misteryEnvelope = [];
+
+ // <FIRST NAME> <LAST NAME> killed Mr.Boddy using the <WEAPON> in the <PLACE
+
+ function revealMistery(misteryEnvelope) {
+
+   return misteryEnvelope[0].first_name+' '+misteryEnvelope[0].last_name+' '+'killed Mr.Boddy using the'+' '+misteryEnvelope[1].name+' in the place '+ misteryEnvelope[2].name;
```
ta-web-paris: It was a problem of index, I change the index so it can match to the sentence : 

```js 

   return misteryEnvelope[0].first_name+' '+misteryEnvelope[0].last_name+' '+'killed Mr.Boddy using the'+' '+misteryEnvelope[2].name+' in the place '+ misteryEnvelope[1].name;


```

starter-code/src/clue.js
```diff
@@ -1,94 +1,192 @@
...
+roomsArray.push(weapon5);
+roomsArray.push(weapon6);
+roomsArray.push(weapon7);
+roomsArray.push(weapon8);
+
+//console.log(roomsArray);
 // Weapons Collection
 var weaponsArray = [];
+
+weaponsArray.push(room1);
```
ta-web-paris: instead of writing the same method like this, just use this simple line and it will do the same : 

```js

weaponsArray.push(room1,room2,room3,room4, ect..) 

```

ta-web-paris:
You can do the same with less code sometimes (as mentionned in my feedback :) ) otherwise you manage to go through all the iteration, your code looks good & it works :)

Keep up the good work!

PR #448, opened 2018-05-30
PAR Meri

mpligioso:


ta-web-paris:
Good job! You manage to go through all the iterations and your code is good :)
There is still some errors left on Jasmine but I can't figure out where they are coming from.. Will take a look at it!

Keep up the good work!

PR #450, opened 2018-05-31
PAR - Kevin Rabesaotra

knandraina:


M-Gadd:
The call backs you have in the last function can be simplified if you access the array directly. 

//
return name.first_name + " " + name.last_name + " killed Mr.Boddy using the " + pickMistery()[1][4]
   .name+ " in the " + pickMistery()[2][5].name+"!!!!"
//

like this ↓

misteryEnvelope[0].first_name
misteryEnvelope[1].name
etc.. 

Good Job

PR #476, opened 2018-06-13
MIA Ryan Soderberg Alexandra Zarrella

Ryanjso:


starter-code/src/clue.js
```diff
@@ -1,94 +1,164 @@
...
+// name: Spa
+// name: Living Room
+// name: Observatory
+// name: Theater
+// name: Guest House
+// name: Patio
+
+
+//Characters
+var characterArray = [
```

sandrabosk:
Good work! Thank you for submitting.

PR #483, opened 2018-06-13
Mia Yaryna Ravlyk WebDev

YokoYolo:
still have one error

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -101,6 +101,7 @@ var roomsArray = [
 ]
 
 
+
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,127 @@
...
-var charactersArray = [];
-
-// Rooms' Collection
-var roomsArray = [];
-
-// Weapons Collection
-var weaponsArray = [];
+
+var charactersArray = 
+[
```

sandrabosk:
Good work, thank you for submitting.

PR #485, opened 2018-06-13
MIA Giannini, Ali, Paola Arrendondo: Now passes all tests! 

GianniniCharles:
MIA WebDev 061118 Giannini, Ali, Paola Arrendondo: Now passes all tests!

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -141,11 +141,10 @@ function randomSelector(anyArray) {
     var i = Math.floor(Math.random() * anyArray.length);
   return anyArray[i];
   }
-var misteryEnvelope = [];
 // ** PICKMYSTERY FUNCTION **
 function pickMistery() {
-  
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,157 @@
...
-name: knife       ---   weight: 8
-name: candlestick ---   weight: 2
-name: dumbbell    ---   weight: 30
-name: poison      ---   weight: 2
-name: axe         ---   weight: 15
-name: bat         ---   weight: 13
-name: trophy      ---   weight: 25
-name: pistol      ---   weight: 20
-
+let weaponsArray = [
```
sandrabosk: This is a bit "workaround" since we are practicing JS methods and this was good chance to create multiple objects and push them into the array (the same as in charactersArray and roomsArray). This works too tho.

sandrabosk:
Very good! Thank you for submitting!

PR #489, opened 2018-06-13
MIA-Andrew Flory

DrewFlory:


starter-code/src/clue.js
```diff
@@ -1,94 +1,65 @@
...
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-occupation:   Entrepreneur
+var charactersArray = [
```

sandrabosk:

PR #545, opened 2018-08-08
PAR Alice & Luca

lucamerzi:
@alicedbrg


starter-code/src/clue.js
```diff
@@ -1,94 +1,144 @@
...
 // Characters Collection
-var charactersArray = [];
+var charactersArray = [
+	mrGreen,
+	drOrchid,
+	profPlum,
+	missScarlet,
+	mrsPeacock,
+	mrMustard
+];
```
ta-web-paris: This way works, but you could also have left the 
```
var charactersArray = [];
```
that was originally in the starter code and added
```
```

ta-web-paris:
Clear and efficient, well done! :100:

PR #546, opened 2018-08-08
Par John & Nerimene

nerimeneakari:
@johndimitroff


ta-web-paris:
Good job! Keep mixing different types of data structures to practice :)
Don't hesitate to come ask if you have any question!

PR #547, opened 2018-08-08
PAR Mohamed & julie 

Youlie03:
@SMoZam

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,205 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-occupation:   Entrepreneur
 
-drOrchid
-frst_name:   Doctor
-lst_name:    Orchid
-color:        white
-description:  PhD in plant toxicology. Adopted daughter of Mr. Boddy
-age:          26
-image:        http://www.radiotimes.com/uploads/images/Original/111967.jpg
-ocupation:   Scientist
+var mrGreen =
+    {
+        first_name:   "Jacob",
+    last_name:    "Green",
+    color:        "green",
+    description:  "He has a lot of connections",
+    age:          45,
+    image:        "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
+    occupation:   "Entrepreneur"
+    };
 
-profPlum
-first_name:   Victor
-last_name:    Plum
-color:        purple
-description:  Billionare video game designer
-age:          22
-image:        https://metrouk2.files.wordpress.com/2016/07/professor-plum.jpg
-occupation:   Designer
+var drOrchid =     {
+    first_name:   "Doctor",
+    last_name:    "Orchid",
+    color:        "white",
+    description:  "PhD in plant toxicology. Adopted daughter of Mr. Boddy",
+    age:          26,
+    image:        "http://www.radiotimes.com/uploads/images/Original/111967.jpg",
+    ocupation:   "Scientist"
+};
 
-missScarlet
-first_name:   Kasandra
-last_name:    Scarlet
-color:        red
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,205 @@
...
+        {
+        first_name:   "Jack",
+        last_name:    "Mustard",
+        color:        "yellow",
+        description:  "He is a former football player who tries to get by on his former glory",
+        age:          62,
+        image:        "https://metrouk2.files.wordpress.com/2016/07/colonel-mustard.jpg",
+        occupation:   "Retired Football player",  
+        }
+];
```
ta-web-paris: Since you created a variable for each character earlier (which is a good call!), you can just push them into the array instead of re-writing everything : 
```
var charactersArray = [];
```

starter-code/src/clue.js
```diff
@@ -1,94 +1,205 @@
...
+    {name: "rope"        ,   weight: 10},
+    {name: "knife"       ,   weight: 8},
+    {name: "candlestick" ,   weight: 2},
+    {name: "dumbbell"  ,   weight: 30},
+    {name: "poison"   ,   weight: 2},
+    {name: "axe"    ,   weight: 15},
+    {name: "bat"     ,   weight: 13},
+    {name: "trophy"      ,   weight: 25},
+    {name: "pistol"      ,   weight: 20}
+];
```
ta-web-paris: Same as before, you can do this : 
```
var weaponsArray = [];
weaponsArray.push(rope, knife, candlestick, dumbbell, poison, axe, bat, trophy, pistol);
```

starter-code/src/clue.js
```diff
@@ -1,94 +1,205 @@
...
+function randomSelector (arrayEntry) {
+    return arrayEntry[Math.floor(Math.random() * arrayEntry.length)];
+};
+
+
+// var misteryEnvelope;
+
+function pickMistery(){
+    
+    misteryEnvelope = [ randomSelector(charactersArray)  , randomSelector(weaponsArray), randomSelector(roomsArray) ];
```
ta-web-paris: You should always declare a variable before using it (don't forget to put var before the first time you use it) : 
```
var misteryEnvelope = [ randomSelector(charactersArray)  , randomSelector(weaponsArray), randomSelector(roomsArray) ];
```
Inside functions, it will work if you don't, but it changes the scope of the var (it becomes global instead of function wide), this is very specific and it is really not advised to do that :)

starter-code/src/clue.js
```diff
@@ -1,94 +1,205 @@
...
+    return  misteryEnvelope;
+    
+};
+
+var misteryEnvelope = pickMistery();
```
ta-web-paris: To explain why you need to write 'var' again here : 
the var misteryEnvelope = pickMistery(); isn't the same one as the one you used earlier inside the pickMistery function : this one has a global scope and is used to store the array returned by the function. The one inside the function has a function scope, you could have called it myMistery or any other name and it would have worked perfectly well :)

starter-code/src/clue.js
```diff
@@ -1,94 +1,205 @@
...
+
+function revealMistery(result){
+result = misteryEnvelope[0].first_name +" "+ misteryEnvelope[0].last_name + " killed Mr.Boddy using the " + misteryEnvelope[1].name + " in the " + misteryEnvelope[2] +"!!!!" ;
+return result;
+}
```
ta-web-paris: one small mistake here : result is what the function returns, so it can't be passed to the function as an argument. You are creating the result variable inside the function, so again you should declare it writing 'var' before. The information the function really needs in order to work is the misteryEnvelope, you should pass this as an argument :)
Right now, when jasmine tests the function,  it tries to pass a certain mistery envelope as an argument but the function currently ignores it and uses the var you created earlier as the array in which it looks for the answer to return, this is why you get an error on Jasmine. Change the parameter and it will be fine !

ta-web-paris:
Variable scopes and the difference between a parameter (abstract thing you use when declaring a function) and an argument (actual variable you use when calling a function) can be tricky, it may be more clear to explain it orally, don't hesitate to come ask if you don't understand my comments or have any extra question!

PR #549, opened 2018-08-08
PAR Aurélien & Constance

Cr27910:
@LOOFFI


starter-code/src/clue.js
```diff
@@ -1,94 +1,171 @@
...
+    myMysteryArr.push(randomSelector (weaponsArray));
+    myMysteryArr.push(randomSelector (roomsArray));
+    return myMysteryArr;
+};
+
+var misteryEnvelope = pickMistery();
+
+function revealMistery (misteryEnvelope){
+    console.log(misteryEnvelope[0].first_name + " " + misteryEnvelope[0].last_name + " killed Mr.Boddy using the " + misteryEnvelope[1].name + " in the " + misteryEnvelope[2].name + " !!!!");
+    return misteryEnvelope;
```
ta-web-paris: Right now, the function is returning the exact thing that has been passed as an argument of the function : an array of three cards  and not a string that reveal the mistery. What you actually want to return is what you console.logged !

starter-code/src/clue.js
```diff
@@ -1,94 +1,171 @@
...
+
+
+
+function pickMistery (){
+    var myMysteryArr = [];
+    myMysteryArr.push(randomSelector (charactersArray));
+    myMysteryArr.push(randomSelector (weaponsArray));
+    myMysteryArr.push(randomSelector (roomsArray));
+    return myMysteryArr;
+};
```
ta-web-paris: To give an explaination about the trouble we had yesterday : 

the    var myMysteryArr = [];   has to be declared inside the function and not outside because otherwise three more cards will be pushed inside the array each time we call the function: 
- three cards are pushed when we first call the function doing var misteryEnvelope = pickMistery();
- when jasmin tests the function, it calls it again, and three more card are pushed to the array, this is why the array length was 6 or more depending on how many times we tried.

right now, the var is declared inside the function and initialised with an empty value : each time we call the function, myMysteryArr is reset to zero, which ensures that only three cards will be in the array after the function does its magic.

PR #561, opened 2018-08-16
MIA FTWD Nick Crispino

crispinonicky:


starter-code/src/clue.js
```diff
@@ -85,10 +98,148 @@ name: Patio
 */
 
 // Characters Collection
-var charactersArray = [];
 
-// Rooms' Collection
-var roomsArray = [];
+var charactersArray = [
```
sandrabosk: The unnecessary creating array of arrays. Characters are just objects so we can push them into the charactersArray.

sandrabosk:

PR #562, opened 2018-08-16
Levijs-MIA-FTWD-clue-exercise

levijskal:


starter-code/src/clue.js
```diff
@@ -92,3 +105,31 @@ var roomsArray = [];
 
 // Weapons Collection
 var weaponsArray = [];
+
+
+var characters = {}
+
+charactersArray.push(mrGreen, drOrchid, profPlum, missScarlet, mrsPeacock, mrMustard)
+
+roomsArray.push(rooms)
```
sandrabosk: This is an unnecessary step, you already have an array so there's no need to push it into an empty array.

starter-code/src/clue.js
```diff
@@ -92,3 +105,31 @@ var roomsArray = [];
...
+    var misteryEnvelope = [
+      randomSelector(charactersArray),
+      randomSelector(weaponsArray),
+      randomSelector(roomsArray)
+    ]
+    return misteryEnvelope
+  }
+
+  function revealMistery(misteryEnvelope) {
+    var simple = `${misteryEnvelope[0].first_name} ${misteryEnvelope[0].last_name} killed Mr.Boddy using the ${misteryEnvelope[1].name} in the ${misteryEnvelope[2].roomNames}!!!!`
```
sandrabosk: Every variable that you create takes part of the memory so try not to create variables when there's no need. You can just:
```
return `${misteryEnvelope[0].first_name} ${misteryEnvelope[0].last_name} killed Mr.Boddy using the ${misteryEnvelope[1].name} in the ${misteryEnvelope[2].roomNames}!!!!`
```

sandrabosk:

PR #570, opened 2018-08-22
finished exercise cluedo Angel and Cyril

cyril-attie:
Difficulties in data format structuring.. we chose to implement an array of strings for rooms and the testing was expecting an array of objects.


ta-web-mad:
Buen trabajo chicos! Quitad los console.logs ya que eso solo es para vosotros! 💪🏽

Anna

PR #616, opened 2018-09-25
Rita Mernissi - lab-javascript-clue

mernri:
Everything passes the tests except pickMistery function. 
I don't understand why it says that my function doesn't return an Array of 3


ZwoRmi:
When I run the tests with your code I see 7 failures, did you push everything ?

PR #617, opened 2018-09-26
Elodie

EloK6:


starter-code/src/clue.js
```diff
@@ -85,10 +85,111 @@ name: Patio
...
+    {name:"Patio"}
+];
+
+function randomSelector(sourceArray) {
+    var keysArr = Object.keys(sourceArray);
+    if (keysArr.length===0) return undefined;
+    var randomIndx = Math.floor(Math.random()* keysArr.length);
+    return sourceArray[keysArr[randomIndx]];
+}
+
```
ta-webpt-par: You need to declare the "var misteryEnvelope" so that it can be defined later on.

starter-code/src/clue.js
```diff
@@ -85,10 +85,111 @@ name: Patio
...
+function pickMistery() {
+    var character = randomSelector(charactersArray);
+    var weapon = randomSelector(weaponsArray);
+    var room = randomSelector(roomsArray);
+    var misteryEnvelope = [character, weapon, room];
```
ta-webpt-par: Since you already declared the variable, you only have to define it without using the "var"

starter-code/src/clue.js
```diff
@@ -85,10 +85,111 @@ name: Patio
...
+    return misteryEnvelope;
+}
+
```
ta-webpt-par: it could be necesary to call on your pickMistery() function using the parameter misteryEnvelope

ta-webpt-par:
Other than that, your code is clean and clear ! Keep it up ! :)

PR #618, opened 2018-09-27
<PAR PT> Sandra

SanDDupt:
Dernier exercice incomplet
@ta-webpt-par


ZwoRmi:
Great work !

There is one test that doesn't pass, because the test expected you to use `first_name` instead of `firstName`

PR #619, opened 2018-09-27
[PAR-PT] Julie Chane Ching

JulieCC:
J'ai tout fait sauf les 3 dernières erreurs que je n'arrive pas à solutionner !
@ta-webpt-par


starter-code/src/clue.js
```diff
@@ -1,94 +1,148 @@
...
-name: Ballroom
-name: Hall
-name: Spa
-name: Living Room
-name: Observatory
-name: Theater
-name: Guest House
-name: Patio
-*/
+function pickMistery () {
```
ta-webpt-par: You need to declare the var misteryEnvelope without defining it oustide of the scope of the function

starter-code/src/clue.js
```diff
@@ -1,94 +1,148 @@
...
+        randomSelector(weaponsArray),
+        randomSelector(roomsArray)
+    ];
+    return misteryEnvelope;
+}
 
-// Characters Collection
-var charactersArray = [];
+function revealmystery(misteryEnvelope) {
+    console.log(
```
ta-webpt-par: don't use the console.log when you want to return results. Either you create a variable that you call in the return or you return the result directly.
For Jasmine to validate your code, the sentence needs to be "exact" (yes, technology is dumb some times ^^)

starter-code/src/clue.js
```diff
@@ -1,94 +1,148 @@
...
+    ;
+    return result;
 
-// Rooms' Collection
-var roomsArray = [];
+}
 
-// Weapons Collection
-var weaponsArray = [];
+console.log(revealmystery(pickMistery()));
```
ta-webpt-par: You just call revealMistery(misteryEnvelope) without the console.log and without uusing a function as a parameter to another function.

PR #621, opened 2018-09-29
work lab-javascript-clue

sylvia-moreno:


ta-webpt-par:
Great work ! Your code is clean !

PR #643, opened 2018-10-17
amine

enima2684:


ta-web-paris:
Perfect !! 
If this is too easy, you can come ask us for some katas to do on codewars to practice more advance stuff !
Also, could you please add 'Paris' before your name when you do a pull request ? it makes us easier to spot which pull requests come from our students when we download the labs to correct :)

PR #650, opened 2018-10-17
Paris - Helen

hwoldu:


ta-web-paris:
Good job on the first part ! Try to practice a bit more and eventually it will become easy :)

PR #658, opened 2018-10-17
Antoine - Clues Exercice

DoneDeal0:


ta-web-paris:
it's good! 

You seem to understand what we've done today in class and why we created the randomSelector function. 

if you finish it, let us know if you want more feedback on it. 👏

PR #659, opened 2018-10-17
ParisPriyanka

priyaprafful:
i didn't find the solution of last function

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,123 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-occupation:   Entrepreneur
 
-drOrchid
-frst_name:   Doctor
-lst_name:    Orchid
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,123 @@
...
+    var randomCharacters = randomSelector (charactersArray);
+    var randomRooms = randomSelector (roomsArray);
+    var randomWeapons = randomSelector (weaponsArray);
+    var misteryArray = [randomCharacters,randomWeapons,randomRooms];
+    return misteryArray;
+}
+
+var misteryEnvelope = pickMistery();
+
+function revealMistery() {
```
ta-web-paris: You have the right idea! You just needed to pass something to your function. I named it myMystery in the example bellow

```
function revealMstery(myMystery) {
  var result =
    myMystery[0].first_name +
    " " +
    myMystery[0].last_name +
    " killed Mr.Boddy using the " +
    myMystery[1].name +
    " in the " +
    myMystery[2].name +
    "!!!!";
  return result;
}
```

You also needed to swap around the order of the last two items in order for it to pass the test. They wanted the person then the weapon and then room.

ta-web-paris:
You were very close to the last solution! 

Great work 👏

PR #660, opened 2018-10-17
Christian Tikombouo

tidochrys:
Here it is .

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -10,8 +10,8 @@ var mrGreen = {
     };
     
     var drOrchid = {
-    frst_name:   "Doctor",
-    lst_name:    "Orchid",
+    first_name:   "Doctor",
+    last_name:    "Orchid",
     color:        "white",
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,154 @@
...
+    {name: "dumbbell", weight: 30},
+    {name: "poison", weight: 2},
+    {name: "axe", weight: 15},
+    {name: "bat", weight: 13},
+    {name: "trophy", weight: 25},
+    {name: "pistol", eight: 20},
+    ];
+    
+    function randomSelector(myArray){
+    var randomItem = myArray[Math.floor(Math.random()*myArray.length)];
```
ta-web-paris: here you don't really have to create a variable if you don't want to. it's ok to just return it on the same line. Of course if you want to add a variable you can 

```
return myArray[Math.floor(Math.random()*myArray.length)]; 
```

ta-web-paris:
Good job! You seem to understand what we covered in todays lesson  💪

PR #670, opened 2018-10-20
[PAR-PT] Saro

ta-webpt-par:
starting the lab
@ta-webpt-par


ta-webpt-par:
It's really great that you're working on the labs to catch on everything ! 
If you change anything, create a new pull so that we can check you're advancement !
GOOD JOB !!
-Jihad

PR #672, opened 2018-10-23
paris-not finished

filippofre:


ta-web-paris:
your code looks good! 💪

You're close to finishing it! I can see you're getting a random character, room and weapon for each time pickMistery() is called. If you finish it, let us know and we'll give you some more feedback :)

PR #682, opened 2018-10-24
AMS - David Z - Finish Cluedo Objects exercise

zlobizlobi:


IronhackAMS:
Nice work! Keep in minds that if you don't use the var keyword when defining an object, its scope will always be global. In this particular use case that's not a problem, but it might be from the moment that you want an object, or variable, to have confined scope.

PR #691, opened 2018-10-24
AMS - Lyudmyla Baron

lxs07a:


IronhackAMS:
Ola Lyudmyla!

Nice work with implementing a switch statement. If you liked this exercise, maybe you can already take a look at object constructors, we gonna work with this concept in a couple of weeks with Mongoose and MongoDB. Each object has a constructor property which can be used as a reference to one central Fundamental Object constructor type. If you're up for it you can try to solve this exercise with that:)
Kido

PR #700, opened 2018-10-24
MIA FTWD Rustam-Samigullin

Rustam1993:


MarcosPalaciosIII:
Great job on the clue lab. Working with test is not easy but it is rather helpful. It helps you to find new ways to figure out your solution which you may not have thought of. At first it can be frustrating, but with time you'll actually learn more thanks to test driven development (TDD).

Keep up the great work.

PR #707, opened 2018-10-24
AMS Ana Kelly

anakellyc:


IronhackAMS:
Very nice work! And nicely done with including some user interaction. Some small remarks. Mind code formatting. The else statements should be on a new line and sometimes indentation is a little bit off. Additionally, look in the object constructors if you want to write the objects more concise. 

Nice work!

PR #710, opened 2018-10-25
AMS Jonathan Chin-Sue

jce200:


starter-code/index.html
```diff
@@ -0,0 +1,39 @@
...
+<script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
+</head>
+
+<body>
+
+<!-- Page Content -->
+<div class="container-fluid h-100">
+  <div class="row h-100 justify-content-center align-items-centerh-100 justify-content-center align-items-center">
+    <div class="col-lg-12 text-center mt-4"> <img src="img/header_logo.png" width="150px" alt=""/>
+      <h1 class="mt-5 playgame" onclick="revealMistery(pickMistery());"><a href="#" class="effect-shine">Who murdered Dr. Black?</a></h1>
```
IronhackAMS: You already have a click event function in your JavaScript. You could also write the revealMistery function in there, so your code is more readable and the js and html files are better separated.

starter-code/src/clue.js
```diff
@@ -1,94 +1,144 @@
...
-name: Ballroom
-name: Hall
-name: Spa
-name: Living Room
-name: Observatory
-name: Theater
-name: Guest House
-name: Patio
-*/
+var mrGreen = {
```
IronhackAMS: If you want an extra challenge see if you can use an object constructor instead of explicitly declaring all the characters one by one.

PR #711, opened 2018-10-25
MIA FTWD Rachel Hofer - JS Clue v1

Rachel-Hofer:
I have not resolved all the errors on this one yet, still working on it. JS takes me FOREVER!


MarcosPalaciosIII:
Good attempt on the clue lab. Working with test is not easy but it is rather helpful. It helps you to find new ways to figure out your solution which you may not have thought of. At first it can be frustrating, but with time you'll actually learn more thanks to test driven development (TDD).

Keep up the good work.

PR #744, opened 2018-11-03
[PTMIA1018] - Chris Rodriguez

Jackslaps:


starter-code/src/clue.js
```diff
@@ -1,94 +1,119 @@
...
 
-mrsPeacock
-first_name:   Eleanor
-last_name:    Peacock
-color:        blue
-description:  She is from a wealthy family and uses her status and money to earn popularity
-age:          36
-image:        https://metrouk2.files.wordpress.com/2016/07/mrs-peacock.jpg
-occupation:   Socialité
+    return Math.floor(Math.random() * arrayName.length);
```
ta-web-mia: This function returns a random number that will be smaller than the length of the array passed as the argument. What you want is for this to return a random element from said array. You're almost there you're just missing one part

- Francisco :)

starter-code/src/clue.js
```diff
@@ -1,94 +1,119 @@
...
-name: knife       ---   weight: 8
-name: candlestick ---   weight: 2
-name: dumbbell    ---   weight: 30
-name: poison      ---   weight: 2
-name: axe         ---   weight: 15
-name: bat         ---   weight: 13
-name: trophy      ---   weight: 25
-name: pistol      ---   weight: 20
+function pickMistery() {
+  let randomSuspect = characters[randomSelector(characters)];
```
ta-web-mia: Once you apply the comment I put above, this will look like: 

let randomSuspect = randomSelector(characters);

PR #748, opened 2018-11-04
Nicolas-Castro

nicolas-castro:
done with labwork for JS Clue


starter-code/src/clue.js
```diff
@@ -1,94 +1,119 @@
...
+  "Theater",
+  "Guest House",
+  "Patio"
+  ]
+ 
 
-// Characters Collection
-var charactersArray = [];
+function randomSelector (someArray){
+  return Math.floor(Math.random() * someArray.length);
```
ta-web-mia: Here you're returning a random number which is shorter than the length of the array being checked. Optimally you'd like to return a random element from the specified array and you'll see how that affects your code in the next comment

starter-code/src/clue.js
```diff
@@ -1,94 +1,119 @@
...
+ }
 
-// Rooms' Collection
-var roomsArray = [];
+ const misteryEnvelope = [];
+ function pickMistery(){
+   const randomSuspect = charactersArray[randomSelector(charactersArray)];
+   const randomWeapon = weaponsArray[randomSelector(weaponsArray)];
```
ta-web-mia: When you apply what I mentioned in the previous comment, this line would change to

const randomWeapon = randomSelector(weaponsArray);

PR #791, opened 2019-01-09
Paris - Raja & Céline

celinedesnoe:
We thought we got everything, we tested the code in real.it and it seems to work :)

However, according to Jasmine, some stuff is not as it would like it to be. 

@raprocode

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,208 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,208 @@
...
+
+function pickMistery() {
+  var threeCards = [];
+  var card1 = randomSelector(charactersArray);
+  // console.log(card1.first_name);
+
+  var card2 = randomSelector(weaponsArray);
+  var card3 = randomSelector(roomsArray);
+
+  threeCards.push(card1.first_name + " " + card1.last_name);
```
ta-web-paris: What you have done is correct, but it doesn't pass the test because it wants you to return a killer etc. By this it means it wants the full object of the killer, weapon and room
```
threeCards.push(card1, card2, card3) 
```
by pushing the full object it passes the test and then you can modify the last part to get it to pass the final test

PR #792, opened 2019-01-09
PARIS - Jules Anoop

julesbadin:
@brwnrcluse


ta-web-paris:
Nice! you passed all the tests, your code is clean and you named all your variables well!
 
Jules, can you please put your initial after your name so we don't accidentally confuse you with the other Jules  🙏

PR #794, opened 2019-01-09
Paris - Christina and Kevin B

CIBailey:
Kevin will push updates later on his end but I had to leave :)

@kevinbarns


starter-code/src/clue.js
```diff
@@ -85,10 +126,61 @@ name: Patio
...
+  "patio"
+];
+
+//Random Selector
+function randomSelector(array) {
+  var random = array[Math.floor(Math.random() * array.length)];
+  return random;
+}
+
+function pickMistery(charactersArray, weaponsArray, roomsArray) {
```
ta-web-paris: here, you don't need to pass anything to the function as a parameter. If you remove them, you pass four more tests 

```
function pickMistery() {
  var christina = [];

  var char = randomSelector(charactersArray);
  var roo = randomSelector(roomsArray);
  var weap = randomSelector(weaponsArray);

  christina.push(char, roo, weap);

  return christina;
}
```

This is because the randomSelector() already takes the array as an parameter and you call this inside this function.

ta-web-paris:
What you've done is good! I'll be able to give you more feedback on the pull request from Kevin :)

PR #800, opened 2019-01-09
Paris - Poeli & Thomas

tomdreamer:
@pbojork

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,161 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-occupation:   Entrepreneur
 
-drOrchid
-first_name:   Doctor
-last_name:    Orchid
-color:        white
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,161 @@
...
+function pickMistery() {
+  var misteryEnvelope = [];
+  misteryEnvelope.push(
+    randomSelector(charactersArray),
+    randomSelector(weaponsArray),
+    randomSelector(roomsArray)
+  );
+  return misteryEnvelope;
+}
+function revealMistery(misteryEnvelope) {
```
ta-web-paris: what you've done here is almost correct, but you just need to make a few changes. 

```
function revealMistery(misteryEnvelope) {
  return (mystery =
    misteryEnvelope[0].first_name +
    misteryEnvelope[0].last_name +
    " killed Mr.Boddy using the " +
    misteryEnvelope[1].name +
    " in the " +
    misteryEnvelope[2].name);
}
console.log(revealMistery(misteryEnvelope));

```


you passed the misteryEnvelope because that already contains what you need (the murderer, weapon and room) so you  don't need to pickMistery again. Here you just needed to return the final result using what you have already worked out 

another small thing- for the weapon you used the same index as for the room

ta-web-paris:
Good job! Your code was good and you passed almost all of the tests. Let me know if you need to me explain the last bit to you- it's hard to explain as a comment

🙌

PR #802, opened 2019-01-10
Paris Aurelie & Jules F

AurelieBeno:
@F-Jules

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,152 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-occupation:   Entrepreneur
 
-drOrchid
-first_name:   Doctor
-last_name:    Orchid
-color:        white
-description:  PhD in plant toxicology. Adopted daughter of Mr. Boddy
-age:          26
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,152 @@
...
+);
+
+// console.log(charactersArray);
+
+function randomSelector(array) {
+  var random = Math.floor(Math.random() * array.length);
+  array.length;
+  return array[random];
+}
+var misteryEnvelope = [
```
ta-web-paris: you're not passing some of the tests because Jasmine wants misteryEnvelope to return an array with three elements 
```
var misteryEnvelope = [
  randomSelector(charactersArray),
  randomSelector(weaponsArray),
  randomSelector(roomsArray)
];
```

starter-code/src/clue.js
```diff
@@ -1,94 +1,152 @@
...
+function pickMistery() {
+  {
+    //  console.log([randomSelector(charactersArray).age])
+
+    return misteryEnvelope;
+  }
+}
+
+function revealMistery() {
+  console.log(
```
ta-web-paris: you need to pass your misteryEnvelope to this function (and when you call it)

You then need to change some of the indexes- the first two should now have the same number (because they're from the same object) and then you need to target specially what you want 
```
misteryEnvelope[0].first_name
```

ta-web-paris:
What you've done is good! If my explanations don't make sense, let me know and i'll explain it to you in person 👏

Abi :)

PR #803, opened 2019-01-10
Paris Kevin & Delphine

Deie:
@kevGhub

Sorry we haven't finished it, but still commit
We'll keep working on it !

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -84,11 +84,206 @@ name: Guest House
 name: Patio
 */
 
-// Characters Collection
-var charactersArray = [];
+var mrGreen = {
+    userName: "mrGreen",
+    firstName: "Jacob",
+    lastName: "Green",
+    color: "green",
+    description: "He has a lot of connections",
+    age: 45,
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -84,11 +84,206 @@ name: Guest House
...
+];
+//randomSelector(weaponsArray);
+
+
+// Functions-------------------------------------------------------
+function randomSelector(cardsArray) {
+    return cardsArray[Math.floor(Math.random() * cardsArray.length)];
+}
+
+function pickMistery(picked) {
```
ta-web-paris: here, you don't need to pass anything to the function. 

For this function to work you need to call your randomSelector() here with the name of an array (eg weaponArray) passed to it. Once you have picked a random item from the array, you need to push that to an array that you'll use later to reveal the mystery 

for example - 

``` 
function pickMistery() {
  var resultArray = [];

  resultArray.push(killer);

}
```

ta-web-paris:
What you've done so far is good! If you'd like any more feedback on it if you continue working on it, or my explanation don't make sense, let me know and i can explain

The goal of labs is not to finish them every time, but to understand what we have covered in class 

Abi  :)

PR #805, opened 2019-01-10
Paris - Francesca & Alexandre

alexjoly13:
@wowfranzik

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -92,3 +92,158 @@ var roomsArray = [];
 
 // Weapons Collection
 var weaponsArray = [];
+
+var mrGreen = {
+  first_name: "Jacob",
+  last_name: "Green",
+  color: "green",
+  description: "He has a lot of connections",
+  age: 45,
+  image:
+    "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
+  occupation: "Entrepreneur"
+};
+
+var drOrchid = {
+  first_name: "Doctor",
+  last_name: "Orchid",
+  color: "white",
+  description: "PhD in plant toxicology. Adopted daughter of Mr. Boddy",
+  age: 26,
+  image: "http://www.radiotimes.com/uploads/images/Original/111967.jpg",
+  ocupation: "Scientist"
+};
+
+var profPlum = {
+  first_name: "Victor",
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -92,3 +92,158 @@ var roomsArray = [];
...
+  hall,
+  spa,
+  livingRoom,
+  observatory,
+  theater,
+  guestHouse,
+  patio
+);
+
+var randomSelector = function(array) {
```
ta-web-paris: your random selector was almost correct - but you need the index of the array  

``` 
  return array[Math.floor(Math.random() * array.length)];
}
```

starter-code/src/clue.js
```diff
@@ -92,3 +92,158 @@ var roomsArray = [];
...
+for (i = 0; i < deck.length; i++) {
+  var pick = randomSelector(deck[i]);
+  mysteryEnvelope.push(deck[i][pick]);
+}
+
+console.log(mysteryEnvelope);
+
+function revealMistery() {
+  console.log(
+    mysteryEnvelope[0].first_name +
```
ta-web-paris: pay attention to your spelling- you've wrote both  `mysteryEnvelope` and `misteryEnvelope` within the same function. Your initial array is spelt `misteryEnvelope` 

You also need to pass misteryEnvelope into your function

starter-code/src/clue.js
```diff
@@ -92,3 +92,158 @@ var roomsArray = [];
...
+);
+
+var randomSelector = function(array) {
+  return Math.floor(Math.random() * array.length);
+};
+
+var deck = [charactersArray, weaponsArray, roomsArray];
+var misteryEnvelope = [];
+
+for (i = 0; i < deck.length; i++) {
```
ta-web-paris: I'm not sure what you're trying to do here- 

in order for the tests to pass you need to write a function called pickMistery(). In this function, you need to call randomSelector() and pass in your different arrays and then push the result into your misteryEnvelope array.  for Example: 

```
function pickMistery() {

 var misteryEnvelope = [];

  misteryEnvelope.push(killer );
  return misteryEnvelope;
}

```

ta-web-paris:
Good job 👏Even though your tests didn't all pass, you were close! let me know if you'd like me to review any of this with you

PR #845, opened 2019-01-29
Jelena-Jukic-Amsterdam

jelenajukic:


starter-code/src/clue.js
```diff
@@ -1,94 +1,253 @@
...
+  color:        "purple",
+  description:  "Billionare video game designer",
+  age:          22,
+  image:        "https://metrouk2.files.wordpress.com/2016/07/professor-plum.jpg",
+  occupation:   "Designer"
+};
+
+
+var missScarlet ={
+
```
lienekechee: Watch out for leaving unnecessary white space (empty lines or spaces).  Typically,  just one line in between declarations, and I would leave none here.

starter-code/src/clue.js
```diff
@@ -1,94 +1,253 @@
...
-name: poison      ---   weight: 2
-name: axe         ---   weight: 15
-name: bat         ---   weight: 13
-name: trophy      ---   weight: 25
-name: pistol      ---   weight: 20
+var rope={
+  name:"rope",
+  weight:10
+};
+
```
lienekechee: Please make sure to always put spaces in between operators.  This is the correct way: 
```
var rope = {
  name: "rope",
  weight: 10
};
```

starter-code/src/clue.js
```diff
@@ -1,94 +1,253 @@
...
+// name: Living Room
+// name: Observatory
+// name: Theater
+// name: Guest House
+// name: Patio
+// */
 
 // Characters Collection
-var charactersArray = [];
+var charactersArray = [mrGreen,drOrchid,profPlum,missScarlet,mrsPeacock,mrMustard];
```
lienekechee: Spaces! 😜

starter-code/src/clue.js
```diff
@@ -1,94 +1,253 @@
...
+
+
+function randomSelector(arr){
+
+   if (arr.length==0){
+     return undefined;
+   }
+
+  var number=arr.length;
+
```
lienekechee: I don't think this variable is necessary; you can just use `arr.length` in your calculation.

PR #846, opened 2019-01-29
Isabela-Dowsley

isabeladowsley:


starter-code/src/clue.js
```diff
@@ -1,94 +1,155 @@
...
+    {name: "Observatory"},
+    {name: "Theater"},
+    {name: "Guest House"},
+    {name: "Patio"},
+];
+
+console.log(roomsArray);
+
+
+
```
lienekechee: Watch out for whitespace :) . Typically one line at most between declarations etc.  Also, be sure not to commit work that is free of debugging code (i.e. `console.log()`)

PR #848, opened 2019-01-29
clue lab solution + frontend

amperry:


starter-code/src/clue.js
```diff
@@ -1,94 +1,258 @@
...
+    name: "Theater"
+}
+
+const guestHouse = {
+    name: "Guest House"
+}
+
+const patio = {
+    name: "Patio"
+}
```
lienekechee: I think I would have declared the rooms a bit differently.  I think naming the Room objects is a little unnecessary because they each have a single property, which is their name.    I would declare the rooms directly in the array: 
```
var rooms = [
{ name: "Guest house" },
{ name: "Patio },
{ name: "Theater" },
{ name: "Observatory" } ]
```
This way you don't have to use the `Array.push` method heaps of times.  Also, you might be interested in checking out constructors.  You can use a constructor function to initialize instances of objects (e.g. Room, Character, Weapon) . Like so:

```
function Room(name) {
  this.name = name;
}
var rooms = [
new Room("Patio"),
new Room("Guest house"),  ...
]
```
But in reality.. totally fine!! I only say this because you might be interested.

starter-code/index.html
```diff
@@ -0,0 +1,29 @@
...
+    <link rel="stylesheet" type="text/css" media="screen" href="styles.css" />
+    <script src="src/clue.js"></script>
+</head>
+<body>
+    <section class="card" id="character">
+        <h1>CHARACTER CARD</h1>
+        <h2 id="character-name" class="name"></h2>
+        <img id="character-pic" src="#" alt="character">
+        <p id="character-desc"></p>
+    </section>
```
lienekechee: Perhaps `character` can be class too. So for your h2: 
```
<h2 class="name character"></h2>
...
<h2 class="name weapon"></h2>
```
and in your css something like:
```
.name:not( .character) {
```

PR #861, opened 2019-01-30
[AMS] - Fedde

Fedduh:
First version. Including a little frontend in HTML.

### DIFF
--- a/starter-code/css/style.css
+++ b/starter-code/css/style.css
@@ -1,0 +1,47 @@
+body {
+  margin: 0px;   
+  background-color: turquoise;
+  background-position: cover;
+  width: 100%; 
+  padding-top: 50px; 
+}
+
+.overview {
+  background-color: turquoise;
+  width: 900px;
+  height: 400px;
+  margin: 0px auto 10px;
+  display: flex;
+  justify-content: space-around;
+}
+
+.block {
+  flex-direction: row;
+  width: 210px;
+  background-color: white;
+  margin: 5px;
+  padding: 5px;
+  border-radius: 20px;
+  }
+
+.block h2 {
+  text-align: center;
+}
+
+.block img {
+  width: 100%; 
+}
+
+.center-container {
+  text-align: center;
+  margin-top: 20px;
+}
+
+.message {
+  padding: 20px;
+  background-color: white; 
+  border-radius: 20px; 
+  display: inline-block;
+  margin: 0px auto 10px;
+}
+ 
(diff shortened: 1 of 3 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,175 @@
...
+console.log(revealMistery(misteryEnvelop));
+
+ 
+ 
+//--------------------//
+// js to HTML
+//--------------------//
+// Character
+document.getElementById("character").src = misteryEnvelop[0].image;
+// document.getElementById("character-name").innerHTML = (misteryEnvelop[0].first_name + " " + misteryEnvelop[0].last_name);
```
lienekechee: Try not to leave comments in code that is committed.   If you want to to keep code that you are not using, try using `git stash` . This command allows you to "stash" all uncommitted changes.  The changes can then be reinstated using `git stash pop` (which reinstates and also deletes the last stash).  Maybe read a little about it on [free code camp](https://medium.freecodecamp.org/useful-tricks-you-might-not-know-about-git-stash-e8a9490f0a1a) :)

starter-code/index.html
```diff
@@ -0,0 +1,63 @@
...
+      <!-- character details -->
+      <div class="block">
+        <h2>Murderer details</h2>
+        <ul>
+          <li id="character-firstname"></li>
+          <li id="character-lastname"></li>
+          <li id="character-age"></li>
+          <li id="character-profession"></li>
+          <li id="character-description"></li>
+        </ul>
```
lienekechee: Generally speaking, if you have to repeat a word several times, it should probably be a class.  If i were you, I would make `character`, `weapon` and `room` classes.    then you could give your block an additional class `character`:
```
  <div class="block character">
        <h2>Murderer details</h2>
        <ul>
          <li id="firstname"></li>
          <li id="lastname"></li>
          <li id="age"></li>
          <li id="profession"></li>
          <li id="description"></li>
        </ul>
```
And then in your JS you could do something like: 
```
var characterName = document.querySelector(".character #firstname")
```

lienekechee: [List of JS DOM Methods](https://www.w3schools.com/jsref/dom_obj_document.asp)

starter-code/css/style.css
```diff
@@ -0,0 +1,47 @@
...
+  background-color: turquoise;
+  width: 900px;
+  height: 400px;
+  margin: 0px auto 10px;
+  display: flex;
+  justify-content: space-around;
+}
+
+.block {
+  flex-direction: row;
```
lienekechee: This I think is the default.

PR #865, opened 2019-01-30
Assignment Day 4

camilav2:


TA-Ironhack-Matt:
Hi Camila and Isabela,

Very good job on the Clue assignment! I really like the way you create the mistery Envelope by calling the randomSelector() inside the array directly. I want to give you a tip so you can make the function even smaller > http://prntscr.com/mevno2 

Keep up the good work:) 

Matt

PR #910, opened 2019-03-20
[MAD PP] Raul Contreras & Anthony S. Batista

anthonysbr:
Comments: It was a bit difficult for my. Thanks to my mate we end it properly. Need to study a bit . more.

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,111 @@
 // Characters
-/*
-mrGreen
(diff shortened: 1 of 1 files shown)


ta-web-mad:
Muy buen trabajo 💪 

Pero tal y como lo habeis subido no pasaba todos los test, he tenido que probar la funcion de  revealMistery() que teniais comentada.

Checked closed Sandra 😸

PR #928, opened 2019-04-03
MIA-FTWD-Allison

allisonkopp:


starter-code/src/clue.js
```diff
@@ -1,94 +1,149 @@
...
+        name: "pistol",
+        weight: 20,
+    },
+]
+
+
+//Rooms
+
+const roomsArray = [
+    { name: "Dinning Room" },
```
MarcosPalaciosIII: when creating arrays, if you use objects that only have 1 key:value pair, it is best to just make it an array with the value.

MarcosPalaciosIII:
Great work on the clue lab. The take away is to try an keep things simple, try not to over think or complicate things.

Keep up the great work.

PR #929, opened 2019-04-03
MIA FTWD - Nour Cherif Esoussi - April 2019

Nourcs:


MarcosPalaciosIII:
Great work on the clue lab. The take away is to try and keep things simple, try not to over think or complicate things.

Keep up the great work.

PR #930, opened 2019-04-03
MIA-FTWB-April Parker Cooper

parkcoop:


starter-code/src/clue.js
```diff
@@ -1,94 +1,191 @@
...
-name: rope        ---   weight: 10
-name: knife       ---   weight: 8
-name: candlestick ---   weight: 2
-name: dumbbell    ---   weight: 30
-name: poison      ---   weight: 2
-name: axe         ---   weight: 15
-name: bat         ---   weight: 13
-name: trophy      ---   weight: 25
-name: pistol      ---   weight: 20
+let roomsArray = [
```
MarcosPalaciosIII: when creating an array, if theres only one key: value pair in the object within the array you should just make it an array of values rather than object.

MarcosPalaciosIII:
Great work on the clue lab. The take away is to try an keep things simple, try not to over think or complicate things. 

Keep up the great work.

PR #934, opened 2019-04-03
MIA FTWD - Angel Biondi

AngelBiondi:
done


MarcosPalaciosIII:
Great work on the clue lab. The take away is to try an keep things simple, try not to over think or complicate things. For instance arrays with objects that have 1 key:value pair can just be an array of values.

Keep up the great work.

PR #936, opened 2019-04-03
MIA FTWD || Samuel Y & Chris C

YamamotoSamuel:


MarcosPalaciosIII:
Great work on the clue lab. The take away is to try and keep things simple, try not to over think or complicate things. Making an array of variables with the object is good but does not look clean. 

Keep up the great work.

PR #938, opened 2019-04-03
MIA FTWD APR-Katy Brown

katybrown91:


MarcosPalaciosIII:
Great work on the clue lab. The take away is to try an keep things simple, try not to over think or complicate things. You don't have to leave commented material. You can use it by modifying it rather than leaving it all there. Or remove unneeded commented material once you have used it.

Keep up the great work.

PR #984, opened 2019-06-05
PP-FT-MX Armando Márquez - Humberto Barajas - José Pablo

jose1987barajas:
This is the finished clue exercise.


ta-web-mex:
Muy buen trabajo, felicidades! :octocat:

PR #987, opened 2019-06-06
[MAD DE] Rubén Vallejo

rubander:
Using the math.generator and math.floor was hard, but for me the hardest thing was the latest property, because you needed to take into account so many values from the previous two that it was very hard to make it work properly even with multiple reviews of the code. There was always a ' or a  {} or something else missing from the code and it gave me errors all the time.


ta-web-mad:
BOOOOMMMM!!!!

PR #997, opened 2019-06-11
MIA-PTWD Fabricio Policarpo

fpolica91:
I had a very hard time understanding what I was supposed to do. Very confusing.


ta-web-mia:
You declared some variables twice, but other than that it looks good!

PR #998, opened 2019-06-11
MIA PTWD Antquanique Dancy

adancy95:
…d created a method to reveal the mystery


ta-web-mia:
Your code looks good! You handled special cases well in your random selector function. Good work!

PR #1013, opened 2019-06-12
MIA - FTWD Matthew Larrubia

mlarrubia:


ta-web-mia:
So I looked into your code, and I have no idea why it's doing this. All of your logic works out, so don't worry too much about this. It's good otherwise!

PR #1014, opened 2019-06-12
MIA-FTWD Steven Englander / Matt Larrubia

confidenti-al:
Completed, but with some errors in one of the test cases for pickMistery when undefined


ta-web-mia:
Jasmine can be kind of weird, so don't get too hung up on those issues. The code is good, otherwise!

PR #1023, opened 2019-06-12
Alexander-Hoinville

ahoin001:
Finished and works, but possible syntax issues not matching failures (Think I tested two different kinds of reveal mistery as well)


ta-web-mia:
It works well. Nice job!

PR #1027, opened 2019-06-13
MIA-FTWD-Manuel Garcia/Serena

manuelga22:
It's weird that my last method works totally fine but I keep getting an error that I can't see the first_name property of undefined but somehow I could when I tested it.


ta-web-mia:
There have been a lot of issues with the last Jasmine problem. Everything else seems to be working, good job!

PR #1060, opened 2019-06-18
PTWD - MIA - Nelly Gonzalez-Vinas

nellygvinas:
I was having a rough time with the Jasmine Specs until it finally dawned on me that the reason was due to the different spellings of the word "Mystery." The lab has it incorrectly spelled as "mistery"!!!

Hopefully this is not a problem for other students, but for grammar geeks it might be lol.

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,110 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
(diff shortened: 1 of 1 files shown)


ta-web-mia:
Excellent work! Hmm, I'll definitely get that test case error fixed. One of the Ironhack team members from Spain made this lab so there are a few spelling errors throughout. Thanks for the heads up :)

PR #1138, opened 2019-08-15
LIS-FTWD-156-KidistCherkos

kidistca:


GTCarmona:
KIDISTTTT

Very good! I got no remarks here so here is your golden star

image


You are doing great, keep this up! We can give you more challenges if you feel like it

PR #1142, opened 2019-08-15
final version of js clue-daniel Santos

DanielJorgeTeloFilipeDosSantos:


soniameller:
Dani, it is vert important that you follow our naming convention for pull requests.
You should type: 
### LIS-FTWD-Daniel Santos #156

In the description/comments you can let us know if you had been struggling with something in particular or if you need feedback!

PR #1143, opened 2019-08-15
LIS-FTWD-GuilhermeSãoPedro #156

guilhermespedro:
hard time to solve "create the mystery" and didn't try iteration 3 but i think i have a clue about what it's needed to be done

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,185 @@
-// Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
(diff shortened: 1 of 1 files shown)


GTCarmona:
Hey Man!

good, I encourage you to try to finish it, once you understand, it is very fast to do it

one thing I want to point out, is that you need to use "," to separate keys inside an object.
and we DON'T use ";" to separate stuff, this is going to break the code.

We only use ";" when we finish our logic, for example at the end of a function of object

Let me know if you decide to go back and finish this one so I can take a look. 

Keep asking questions and engaged, you are doing well !!!

PR #1146, opened 2019-08-15
LIS-FTWD-[GUSTAVO] #156

gusmagnago:
95% complete

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -179,28 +179,10 @@ const randomSelector = (array) => {
     }; return mystery
   };  
 
-  const revealMystery = (mysteryEnvelope) => {
-    [randomCharacter, randomWeapon, randomRoom];
-    for (let mysteryEnvelope of revealMystery) {
-      `${randomCharacter},"killed Mr. Boddy using the", ${randomWeapon}, "in the", ${randomRoom}`
-    } for (let)
-  } return mysteryEnvelope
+  const revealMystery = () => {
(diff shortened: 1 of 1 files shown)


GTCarmona:
hey Gustavo!!!

Good job! I notice that you probably bang your head not understanding what was going on the last test

I changed it up a little bit and pass the test, like this:
```
   const revealMystery = (pickMystery) => {

    return pickMystery.suspect.first_name + " " + pickMystery.suspect.last_name + " killed Mr. Boddy using the " + pickMystery.weapon.name + " in the " + pickMystery.room.name+"!!!!"
   
  } 
```

Let's compare with what you had and analyze what I did!
```
  const revealMystery = () => {
    const mysteryEnvelope = pickMystery()

  return mysteryEnvelope.suspect.first_name + mysteryEnvelope.suspect.last_name + " killed Mr. Boddy using the" + mysteryEnvelope.weapon.name+ "in the" + mysteryEnvelope.room.name+"!!!!"
 
} 
```

First thing, you created a variable const misteryEnvelope, however you already have a perfectly fine function for that, so you can just use it as the argument of your function

then, you really neeeeeeeed to have everything the same as Jasmine asks, and I can point out you were not using proper spaces between words. Good way to check is using Jasmine itself, to see what are you output, and what is Jasmine expecting. like this:

image

So I just added some spaces between words and now is fine. Make sure you use the debugging tools when you face roadblocks like this! Keep up the good work

PR #1147, opened 2019-08-15
LIS-FTWD-Elyas-#156

elyasp:
Third iteration in progress. Could not figure out the part of
let mysteryEnvelope = "";

test accepted it but I did not pushed the object there.??

### DIFF
--- a/starter-code/.vscode/settings.json
+++ b/starter-code/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -175,6 +175,7 @@ const pickMystery = () => {
   }
 
 
-  //const revealMystery = (mysteryEnvelope) => {
-      // return
-  //}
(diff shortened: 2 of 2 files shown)


soniameller:
Hey Elyas!

I think it is just an instruction problem because you have it already solved with an incorrect name. To be more clear, you were asked to:

### **Create the mystery**
We need to create a pickMystery function that will call randomSelector for each card stack, and return an object with the 3 picked cards, a suspect, a weapon and a room. **Our mystery should be stored on a mysteryEnvelope variable.**

--- 

This is your code:
```
const pickMystery = () => {
    const randomCharacter = randomSelector(charactersArray)
    const randomWeapon = randomSelector(weaponsArray)
    const randomRoom = randomSelector(roomsArray)
    let mystery = {
        weapon : randomWeapon,
        room : randomRoom,
        suspect : randomCharacter 
    }
    return mystery
}

```

You are already creating an object with the 3 picked cards and you are naming it `mystery` while the instructions are asking you to name it mysteryEnvelope. That's it 🙃

PR #1148, opened 2019-08-15
LIS-FTWD-AndreAlmeida #156

andreH2M:


GTCarmona:
Hey Andre! I want to help you fix some small syntax error that are probably breaking your code and not allowing you to move forward:


### Using the console on chrome

You started with the right logic. One thing I really want to point out is, when we are dealing of objects, we should add the "," at the end to separate them! Make sure you use the Chrome console to check for this kind of error, it points it out where is the mistake happening

image

You can identify that on line 50 of your clue.js file, something is breaking. When analyzing the object, we can see the missing ","
image



### Suspects format

```
let mrGreen = 
{
"id": "mrGreen", // You are missing a "," here at the end of the id value!
"first_name": "Jacob",
"last_name": "Green",
"color": "green",
"description": "He has a lot of connections",
"age": 45,
"image": "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
"occupation": "Entrepreneur"}
```

### Weapons array

You should create a unique object for every different weapons and rooms, example
```
const rope = {
    "name": "rope",
    "weight": 10
    };
```

### Rooms Array

Another thing I noticed is that you were defining the array like this:
```
const = roomsArray [

"name": "Dining Room"
"name": "Conservatory"
......
]

```
This is going to point out a mistake, we should define it like this:

```
const roomsArray = [

"name": "Dining Room",
"name": "Conservatory",
......
]

```


Im not going to trough the actual functions and tests, because I don't think you manage to get into  it. But I highly encoure you to try to finish it, so you consolidate the logic of creating and calling functions.


Don't hesitate to call us out, we are here to support. Remember, don't be afraid of asking the basic questions in class, nobody will judge your and we all will get to the finish lane together

PR #1187, opened 2019-08-23
Clue game mystery pick done

JakobPri:


Cule219:
You have a few unneeded declaration at the end, also you shouldn't invoke your methods.
But good job!

PR #1188, opened 2019-08-23
[noRef] mysolution

pintocr:
Hello Ironhack Team,

here my solution

Cristiano


Cule219:
Your revealMystery function needs to look for properties of an object you are trying to access elements of matrix array instead.
Good job other than that!

PR #1191, opened 2019-08-23
done

IliesBourib-mms:


Cule219:
Your pickMystery function needs to return an object with properties. Other than that good job

PR #1250, opened 2019-10-16
[Berlin-WDFT-Daniel and William]

whbrown:
jasmine killed us, last spec fails for no reason even though it works


whbrown:
fixed the last spec for jasmine

PR #1253, opened 2019-10-16
[BER_FTWD_NINO_Hanna]

Lorelore22:


starter-code/src/clue.js
```diff
@@ -1,94 +1,214 @@
...
+
+const patio = {
+    name: "Patio"
+};
+
+
 
 // Characters Collection
 const charactersArray = [];
+charactersArray.push("mrGreen", "drOrchid", "profPlum", "missScarlet", "mrsPeacock", "mrMustard");
```
Svemakawe: Here you are pushing strings inside the array, not the actual variable that withholds the object as a value. The Jasmin Test is just passed , cause its not actually using those arrays. So you're passing the test. But normally this would not work;)..so push it without quotes.

PR #1254, opened 2019-10-16
[BER-WDFT-102019 Noubar Michael]

MikeVanYarnrider:
Done

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,191 @@
 // Characters
-/*
-mrGreen
-first_name:   Jacob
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,191 @@
...
 // Characters Collection
-const charactersArray = [];
+const charactersArray = [
+  mrGreen,
+  drOrchid,
+  profPlum,
+  missScarlet,
+  mrsPeacock,
+  mrMustard
+];
```
Svemakawe: You could've also used the push method here

PR #1255, opened 2019-10-16
[BER-FTWD-Jeff-Fernando]

Jeff-Moraes:


starter-code/src/clue.js
```diff
@@ -85,10 +85,122 @@ name: Patio
...
+  "Lounge",
+  "Ballroom",
+  "Hall",
+  "Spa",
+  "Living Room",
+  "Observatory",
+  "Theater",
+  "Guest House",
+  "Patio"
+];
```
Svemakawe: So here you were supposed to create an array with OBJECTS inside. So each room object has a key "name". Otherwise you couldn't access mysteryEnvelope.room.name in your last iteration. Jasmin is not testing this last function with the arrays you created, instead "she" uses her own, so thats why you still pass the test;)

PR #1256, opened 2019-10-16
BER_FTWD_Shan_Pascal

PascalH91:


starter-code/src/clue.js
```diff
@@ -82,13 +123,98 @@ name: Observatory
...
+const GuestHouse = {
+    name: "Guest House"
+};
+const Patio = {
+    name: "Patio"
+};
 
 // Characters Collection
-const charactersArray = [];
+const charactersArray = [mrGreen, drOrchid, profPlum, missScarlet, mrsPeacock, mrMustard];
```
Svemakawe: you could've also used the push method here: charactersArray.push(mrGreen, drOrchid...)

starter-code/src/clue.js
```diff
@@ -82,13 +123,98 @@ name: Observatory
 name: Theater
 name: Guest House
 name: Patio
-*/
+ */
+
+
+const DiningRoom = {
+    name: "Dining Room"
+};
```
Svemakawe: only class-variables are written Uppercased, variables like this one should start lowerCased-> const diningRoom
..not breaking your code here, but looks confusing cause someone might expect a class:)

PR #1257, opened 2019-10-16
[BER-FTWD-102019] - Alex and Are

arews:


starter-code/src/clue.js
```diff
@@ -1,73 +1,130 @@
...
+firstName:   "Jack",
+lastName:    "Mustard",
+color:        "yellow",
+description:  "He is a former football player who tries to get by on his former glory",
+age:          62,
+image:        "https://static.independent.co.uk/s3fs-public/thumbnails/image/2016/07/04/08/unspecified-3.jpg",
+occupation:   "Retired Football player"
+};
+
+let charactersArray = [mrGreen, drOrchid, profPlum, missScarlet, mrsPeacock, mrMustard];
```
Svemakawe: you could've also used the push-method here. 
`let charactersArray`
`charactersArray.push(mrGreen,drOrchid...)`

starter-code/src/clue.js
```diff
@@ -1,73 +1,130 @@
...
 name: bat         ---   weight: 13
 name: trophy      ---   weight: 25
-name: pistol      ---   weight: 20
+name: pistol      ---   weight: 20*/
 
 // Rooms
-name: Dining Room
+
+const roomsArray = ["Dining Room", "Conservatory", "Kitchen", "Study", "Library", "Billiard Room", "Lounge", "Ballroom", "Hall", "Spa", "Living Room", "Observatory", "Theater", "Guest House", "Patio"]
+
```
Svemakawe: So here you were supposed to create an array with OBJECTS inside. So each room object has a key "name". Otherwise you couldn't access mysteryEnvelope.room.name in your last iteration. Jasmin is not testing this last function with the arrays you created, instead "she" uses her own, so thats why you still pass the test;)

PR #1258, opened 2019-10-16
done - Berlin WebDev FT - Amelia + Laura

ning3l:


starter-code/src/clue.js
```diff
@@ -1,94 +1,173 @@
...
-name: rope        ---   weight: 10
-name: knife       ---   weight: 8
-name: candlestick ---   weight: 2
-name: dumbbell    ---   weight: 30
-name: poison      ---   weight: 2
-name: axe         ---   weight: 15
-name: bat         ---   weight: 13
-name: trophy      ---   weight: 25
-name: pistol      ---   weight: 20
+const weaponsArray = [{
```
itstheandre: even though this is clearly working, these objects have no variable associated to them. For you to know what rope is, you would have to go to weaponsArray[0]. Ideally you would create a variable (rope) and add those properties. And then in the array just put the variable in.

PR #1260, opened 2019-10-16
BER-wdft-Kyle_and_Madhavi

bleepblopblorpblopbloop:


DalinaLeila:
Nice work! 💯 I like the use of ~~ 🤓. When you return your mysteryString you might want to change it to the ES6 (Using backticks `` and the ${} sign) notation to make your code cleaner.

PR #1261, opened 2019-10-16
[BER-FTWD-Stephen-Johanna

scoley89:
lab-javascript-clue

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,94 +1,220 @@
-// Characters
-/*
-mrGreen
-first_name:   Jacob
-last_name:    Green
-color:        green
-description:  He has a lot of connections
-age:          45
-image:        https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg
-occupation:   Entrepreneur
+const mrGreen = {
+  first_name: "Jacob",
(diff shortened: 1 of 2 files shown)


starter-code/src/clue.js
```diff
@@ -1,94 +1,220 @@
...
+  name: "GuestHouse"
+};
+const Patio = {
+  name: "Patio"
+};
 
 // Characters Collection
 const charactersArray = [];
-
+charactersArray.push(
```
itstheandre: You could have just added these elements directly int he array, instead of pushing them into the array.

starter-code/src/clue.js
```diff
@@ -1,94 +1,220 @@
...
+  mysteryEnvelope.weapon = randomSelector(weaponsArray);
+  mysteryEnvelope.room = randomSelector(roomsArray);
+  mysteryEnvelope.suspect = randomSelector(charactersArray);
+  console.log(mysteryEnvelope);
+  return mysteryEnvelope;
+}
+
+function revealMystery(mysteryEnvelope) {
+  return (
+    mysteryEnvelope.suspect.first_name +
```
itstheandre: Here you could have also written with template literals - `` :)

PR #1262, opened 2019-10-16
Berlin-FTWD-Fatma/Pavlina

Pavlitsa:
Completed


DalinaLeila:
Nice work! 💯  Have a look at the use of let and const in your code. In this exercise the predefined content will not change, so you can use const when naming all your variables. Also when you return your final mystery string, you can use ES6 notation to make your code cleaner.

PR #1263, opened 2019-10-16
BER-FTWD-Patrick-Virginia

Vi-GR:


DalinaLeila:
Nice work! 💯 Have a look at the use of let and const in your code. In this exercise the predefined content will not be altered, so you can use const throughout when naming your variables. Your weapon and your character array look great, you could have structured the rooms array in a similar way - This would have saved you some time 🤓. Finally, when you return your mistery string you can use the ES6 notation to make your code cleaner.

PR #1276, opened 2019-10-23
MIA FTWD OCT Nico Bared

nbared:
first submission, having trouble with the revealMystery function


ta-web-mia:
Still needs work on passing the last jasmine tests. Let me know if you need help with the revealMystery function.

PR #1279, opened 2019-10-23
[BCN-FT-Rose Beatty]

Rosebeatty:


starter-code/src/clue.js
```diff
@@ -1,94 +1,213 @@
...
+console.log(roomsArray);
+console.log(weaponsArray);
+*/
+
+function randomSelector (arrayType) {
+    var randomElement = arrayType[Math.floor(Math.random() * arrayType.length)];
+        return randomElement;
+}
+
+var mysteryEnvelope = new Object()
```
tawebbcn: This is perfectly correct but you can also declare a new empty object like **var myNewObject = { }**

PR #1332, opened 2019-10-31
[PTMIA102019] Michel Maia

dimermichel:
Done


MarcosPalaciosIII:
Great work on the clue lab. Creating function to run small tasks will help to keep your code clean.
This also makes it easier to manipulate your code without possibly messing up your function as well as making all those functions reusable if you would like to call them from within additional functions without having to rewrite your code over in another location.

Keep up the great work.

PR #1389, opened 2020-01-08
BER-FT-0120 Victoria-Vaso

vasilikibalidou:


starter-code/src/clue.js
```diff
@@ -1,14 +1,118 @@
...
+  { name: "axe", weight: 15 },
+  { name: "bat", weight: 13 },
+  { name: "trophy", weight: 25 },
+  { name: "pistol", weight: 20 }
+];
 
 // ITERATION 2
 
+function selectRandom(arr) {
+  let num = Math.floor(Math.random() * arr.length);
```
DalinaLeila: can be reduced to 1 line:
`return arr[Math.floor(Math.random() * arr.length)]`

PR #1390, opened 2020-01-08
[BER-FT]Karolina-Sebastian

MngSebastian:


starter-code/src/clue.js
```diff
@@ -1,14 +1,125 @@
...
+  { name: "poison", weight: 2 },
+  { name: "axe", weight: 15 },
+  { name: "bat", weight: 13 },
+  { name: "trophy", weight: 25 },
+  { name: "pistol", weight: 20 }
+];
 
 // ITERATION 2
+function selectRandom(array) {
+  let randomNumber = Math.floor(Math.random() * array.length);
```
DalinaLeila: can be written in 1 line:
`return array[Math.floor(Math.random() * array.length)]`

PR #1391, opened 2020-01-08
[BER FT] Kiyohide + Nora

Norizinus:


starter-code/src/clue.js
```diff
@@ -1,14 +1,185 @@
...
+  {
+    name: "Patio"
+  }
+];
 
 // ITERATION 2
 
+function selectRandom(cardStack) {
+  let card;
+  if (cardStack !== []) {
```
itstheandre: In this case, for a cleaner code, you could check if it's equal to an empty array by cardStack.length === 0

PR #1394, opened 2020-01-08
[BER-FT-012020] Mia & Ciwan

ciwan-aram:
:)

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,14 +1,139 @@
 // ITERATION 1
 
+const mrGreen = {
+  firstName: "Jacob",
+  lastName: "Green<br",
+  occupation: "Entrepreneur",
+  age: 45,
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -1,14 +1,139 @@
...
+function revealMystery(envelope) {
+  return (
+    envelope.suspect.firstName +
+    " " +
+    envelope.suspect.lastName +
+    " killed Mr. Boddy using the " +
+    envelope.weapon.name +
+    " in the " +
+    envelope.room.name +
+    "!"
```
itstheandre: In this case you could have used template literals for an easier to read code:
```
return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}`
```

PR #1395, opened 2020-01-08
[BER-FT] Tim Martin

martinschmal:


starter-code/src/clue.js
```diff
@@ -1,14 +1,186 @@
...
+	{
+		name: "pistol",
+		weight: 20
+	}
+];
 
 // ITERATION 2
 
+function selectRandom(suspectsArray) {
+	let suspect = suspectsArray[Math.floor(Math.random() * suspectsArray.length)];
```
DalinaLeila: could even do 1 line only here:

`return suspectsArray[Math.floor(Math.random() * suspectsArray.length)];`

starter-code/src/clue.js
```diff
@@ -1,14 +1,186 @@
...
+}
+let suspect = "";
+
+function pickMystery() {
+	let weapon = selectRandom(weaponsArray);
+	let room = selectRandom(roomsArray);
+
+	let suspect = selectRandom(suspectsArray);
+
+	let mystery = { weapon: weapon, suspect: suspect, room: room };
```
DalinaLeila: can be reduced to 

`let mystery = { weapon, suspect, room };`

starter-code/src/clue.js
```diff
@@ -1,14 +1,186 @@
...
+}
 // ITERATION 3
+
+function revealMystery(mystery) {
+	let last = mystery.suspect.lastName;
+	let first = mystery.suspect.firstName;
+	let weapon = mystery.weapon.name;
+	let location = mystery.room.name;
+
+	x = first + " " + last + " killed Mr. Boddy using the " + weapon + " in the " + location + "!";
```
DalinaLeila: use backticks and $ sign. Will make your code cleaner :) Also think about how you name your variables. Maybe e.g.`output` would be better than `x`. Also make sure to declare it using var, const or let.

```
const output = `${first} ${last} killed Mr. Boddy using the ${weapon} in the ${location}!`
```

starter-code/src/clue.js
```diff
@@ -1,14 +1,186 @@
...
+	toString(x);
```
DalinaLeila: then you can get rid of everything below and simply

`return output`

PR #1400, opened 2020-01-08
[BER - FT] Laurel & Menna

lwrocks:
Done - would love to review/get feedback for the final bit of code for function revealMystery().

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,14 +1,163 @@
 // ITERATION 1
 
 // Suspects Collection
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: "Jacob",
+    lastName: "Green",
+    occupation: "Entrepreneur",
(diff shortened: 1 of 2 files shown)


starter-code/src/clue.js
```diff
@@ -1,14 +1,163 @@
...
+}
+
+
+
 
 // ITERATION 3
+function revealMystery(inputObj) {
+    let firstName = inputObj.suspect.firstName;
+    let lastName = inputObj.suspect.lastName;
+    let weapon = inputObj.weapons;
```
fonsogms: to acces to the right valeu you can use `inputObj.weapon.name`

starter-code/src/clue.js
```diff
@@ -1,14 +1,163 @@
...
+    let room = inputObj.rooms;
```
fonsogms: Same for room `inputObj.room.name`

PR #1433, opened 2020-01-14
[BCN-WEB-HENRIETTE] Done

Jethet:


starter-code/src/clue.js
```diff
@@ -1,14 +1,200 @@
...
+    mrMustard = {
+        firstName: "Jack",
+        lastName: "Mustard",
+        occupation: "Retired Football player",
+        age: 62,
+        description: "He is a former football player who tries to get by on his former glory",
+        image: "https://static.independent.co.uk/s3fs-public/thumbnails/image/2016/07/04/08/unspecified-3.jpg",
+        color: "yellow"
+    }
+];
```
tawebbcn: You can't declare or initialize a variable inside an array (here you're initializing mrGreen, mrMustard etc, inside the suspectsArray). In that case it works, because JS is a weakly typed language and thus is quite permissive, but if you try to declare the variables with **var** mrGreen = {} inside the array, you'll see that you get an error. 
What you could do is declare each object outside the array and then push them to the suspectArray, or just add the suspects objects without assigning them an identifier.

starter-code/src/clue.js
```diff
@@ -1,14 +1,200 @@
...
+    {
+    name: "pistol",
+    weight: 20
+    }
+];
 
 // ITERATION 2
 
+var card;
+
```
tawebbcn: this global variable is not used, you could remove it (you already declare a local variable with the same name inside the function selectRandom())

PR #1434, opened 2020-01-14
[WEB-BCN-VALENTIN]

valentin-muller:


tawebbcn:
Really good job, good logic 👍

PR #1440, opened 2020-01-14
[BCN -WDFT- RODRIGO MARTINEZ - HOMEWORK DAY 2]

rodmartinezmedina:


starter-code/src/clue.js
```diff
@@ -1,14 +1,165 @@
...
+}
+
 // ITERATION 3
+
+var mysteryEnvelope = pickMystery();
+
+function revealMystery (envelope) {
+  return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`
+};
+
```
tawebbcn: Here 'envelope.room.name' returns undefined because in the roomsArray, you didn't use any key 'name' to store the values of each room. So you would just need to write ${envelope.room} ;)

PR #1443, opened 2020-01-14
[BCN-WDFTJAN2020-JORGE&AKAPON]

jorgeberrizbeitia:


tawebbcn:
Very good job! Clean code and very understandable logic. The only tip i can give you is that when working with more complex project you will have to commit with more meaningful messages but for this project is okay. Good job

PR #1444, opened 2020-01-14
[FT-BCN-012020]-Tarek.Koushaji

Tking1606:
Clue Lab is done

### DIFF
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -92,5 +92,23 @@ const weaponsArray = [
 ];
 
 // ITERATION 2
+function selectRandom(arr){
+ return arr[Math.floor(Math.random() * arr.length)];
 
+}
+function pickMystery(){
+    
+    
+    var x={
+        suspect :selectRandom.suspectsArray,
+        room : selectRandom.roomsArray,
+        weapon : selectRandom.weaponsArray
+    }
+    return x;
+}
 // ITERATION 3
+function revealMystery(obj){
+   return `${obj.suspect.firstName} ${obj.suspect.lastName} killed Mr. Boody 
+   using the ${obj.weapon.name} in the ${obj.room.name}`
+
+}
--- a/starter-code/src/clue.js
+++ b/starter-code/src/clue.js
@@ -1,13 +1,95 @@
 // ITERATION 1
 
 // Suspects Collection
-const suspectsArray = [];
+const suspectsArray = [
+    {
+     firstName: 'Jacob',
+     lastName: 'Green',
+     occupation: 'Entrepreneur',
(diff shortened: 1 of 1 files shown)


starter-code/src/clue.js
```diff
@@ -92,5 +92,23 @@ const weaponsArray = [
...
 // ITERATION 2
+function selectRandom(arr){
+ return arr[Math.floor(Math.random() * arr.length)];
 
+}
+function pickMystery(){
+    
+    
+    var x={
+        suspect :selectRandom.suspectsArray,
```
tawebbcn: This returns undefined. To get a random item it should have been something like suspect:selectRandom(suspectsArray)

starter-code/src/clue.js
```diff
@@ -92,5 +92,23 @@ const weaponsArray = [
...
+        room : selectRandom.roomsArray,
+        weapon : selectRandom.weaponsArray
+    }
+    return x;
+}
 // ITERATION 3
+function revealMystery(obj){
```
tawebbcn: this also returns undefined because of the prevous issue

starter-code/src/clue.js
```diff
@@ -1,13 +1,95 @@
...
 // Suspects Collection
-const suspectsArray = [];
+const suspectsArray = [
+    {
+     firstName: 'Jacob',
+     lastName: 'Green',
+     occupation: 'Entrepreneur',
+     age: 45,
+     description: 'He has a lot of connections',
+     image: mrGreenImage.src='//pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg',
```
tawebbcn: this throws an error "mrgreenImage is not defined". I think you might have confused the image property of this particular object with an HTML tag. It was supposed to be:

image: '//pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg',

tawebbcn:
good job overall. You had some mistakes that will improve with a little more practice but you're doing very well, Tarek. Keep it up! :)

starter-code/src/clue.js
Tking1606: I tried this and it was still undefined ....

starter-code/src/clue.js
Tking1606: that was a mistake ... I wasn't that focused :)

starter-code/src/clue.js
Tking1606: > good job overall. You had some mistakes that will improve with a little more practice but you're doing very well, Tarek. Keep it up! :)

really thanks ..... these comments cheers me up every time I read them XD

PR #1446, opened 2020-01-16
Done

danielhoogerwerf:


starter-code/src/clue.js
```diff
@@ -1,14 +1,115 @@
...
+    return {
+			suspect: selectRandom(suspectsArray),
+			weapon: selectRandom(roomsArray),
+			room: selectRandom(roomsArray)
+		};
+}
+
+const finalSuspect = pickMystery();
+
+const revealMystery = () => (`${finalSuspect.suspect.firstName} ${finalSuspect.suspect.lastName} killed Mr. Boddy using the ${finalSuspect.weapon.name} in the ${finalSuspect.room.name}!`);
```
lotofcaffeine: Really good that you are using Arrow functions. 👏🏻

My only suggestion is next time you could try to isolate your function. 
Example: 
`const revealMystery = (envelope) => (`${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`);`
This way `revealMystery` works regardless of the previous call of `PickMystery`

danielhoogerwerf: Thanks!!

PR #1449, opened 2020-01-16
Done! [AMS-PT-012020] - Alex

alexpanko:


starter-code/src/clue.js
```diff
@@ -1,14 +1,119 @@
...
+    {name: 'Observatory'},
+    {name: 'Theater'},
+    {name: 'Guest House'},
+    {name: 'Patio'}
+]
 
 // ITERATION 2
+function selectRandom(array) {
+    let i = Math.floor(Math.random()*array.length);
+    return array[i]
```
LloydC: You can also just write return array[Math.floor(Math.random()*array.length)];

PR #1503, opened 2020-01-22
[LIS JAN 20 - Matheus Carvalho]

carvalhobfr:
@ironhacklis-webTeam
I think it's right but the specrunner is not completing.


ironhackLis-webTeam:
Looks good. Which of all the test is not passing with Jasmine?

PR #1527, opened 2020-02-04
BER WD PT Sjard and Pia

Musikantensjardl:


starter-code/src/clue.js
```diff
@@ -1,14 +1,102 @@
...
+randomNumber = 0; 
+function selectRandom(type) {
+    randomNumber = Math.floor(Math.random() * type.length)
+return type[randomNumber]
+}
+selectRandom(suspectsArray)
+console.log()
+
 
 // ITERATION 3
```
numbing: so far so good but please try to the next iteration as well, it will help you to learn even more

numbing: and let me know if you need help

PR #1532, opened 2020-02-04
Ber WDPT Camille

camilleory:


starter-code/src/clue.js
```diff
@@ -1,14 +1,161 @@
...
+
+selectRandom(suspectsArray)
+let mystery = {}
+
+function pickMystery() {
+    mystery.suspect = selectRandom(suspectsArray)
+    mystery.weapon = selectRandom(weaponsArray)
+    mystery.room = selectRandom(roomsArray)
+    console.log(mystery)
+    return mystery
```
numbing: it better to bring mystery variable inside the function, its just cleaner. nice job well done

PR #1599, opened 2020-03-25
BER-FTWD-Manuel-Phillip

phpaul89:


src/clue.js
```diff
@@ -1,14 +1,145 @@
...
+        name: "pistol",
+        weight: 20
+    }
+];
 
 // ITERATION 2
 
+function selectRandom(array){
+    const randomElement = array[Math.floor(Math.random() * array.length)];
+    return randomElement;
```

PR #1600, opened 2020-03-25
berlin-ftwd-timur-sebastian

SebastianNagl:


src/clue.js
```diff
@@ -1,14 +1,126 @@
 // ITERATION 1
 
+//randomize arr length to pick rnd
+// let testArr = [12,23,34]
+
+let selectRandom = (arr) => {
+
+    if (arr.length === 0) {
+        return undefined;
+    } else {
```
fenrew: Since you are returning within the first if statement, you don't have to make an else statement here

src/clue.js
```diff
@@ -1,14 +1,126 @@
...
+    //the rnd item picked
+    let randomItem = arr[Math.floor(Math.random() * arr.length)]
+    return randomItem;
```
return arr[Math.floor(Math.random() * arr.length)]

PR #1601, opened 2020-03-25
BER-FTWD-Avtandil-Agustina

agusgutierrezz:


src/clue.js
```diff
@@ -1,14 +1,139 @@
...
+    }
+];
 
 // ITERATION 2
-
+function selectRandom(array){
+    return array[Math.floor(Math.random() * array.length)];
+}
+function pickMystery(){
+    const newObject = {
```
fenrew: You dont have to create an object like this actually. You can rather just write:
const newObject = {}

PR #1602, opened 2020-03-25
BER-FTWD-Lucas+Yosra

yosrayousfi:
All done.


src/clue.js
```diff
@@ -1,14 +1,150 @@
...
+    name: "pumpgun",
+    weight: 50
+  }
+];
 
 // ITERATION 2
+function selectRandom(array) {
+  let randomlySelected = 0;
+  randomlySelected = array[Math.floor(Math.random() * array.length)];
+  return randomlySelected;
```

PR #1609, opened 2020-03-25
BER-FTWD-Victoria&Olha

olhadanylchenko:
DOOOOONE

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,13 +1,143 @@
 // ITERATION 1
+function selectRandom(arr) {
+  return arr[Math.floor(Math.random() * Math.floor(arr.length))];
+}
+function pickMystery() {
+  const objectWithCards = {
+    suspect: selectRandom(suspectsArray),
+    weapon: selectRandom(weaponsArray),
+    room: selectRandom(roomsArray)
+  };
+  return objectWithCards;
+}
+
+function revealMystery(envelope) {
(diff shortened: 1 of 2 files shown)


src/clue.js
```diff
@@ -1,13 +1,143 @@
 // ITERATION 1
+function selectRandom(arr) {
+  return arr[Math.floor(Math.random() * Math.floor(arr.length))];
+}
```
Jeff-Moraes: nice approach!
You could also take the last Mat.floor out because the arr.length will already return an integer number.

src/clue.js
```diff
@@ -1,13 +1,143 @@
...
+function pickMystery() {
+  const objectWithCards = {
+    suspect: selectRandom(suspectsArray),
+    weapon: selectRandom(weaponsArray),
+    room: selectRandom(roomsArray)
+  };
+  return objectWithCards;
+}
```
Jeff-Moraes: You don't have to create an object like this actually. You can rather just write:
`const newObject = {}`
and then assign the new proprietors to it.

src/clue.js
```diff
@@ -1,13 +1,143 @@
...
+
+function revealMystery(envelope) {
+  let message = `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`;
+  return message;
+}
```
Jeff-Moraes: nicely done. ;)

PR #1610, opened 2020-03-25
Berlin ft aleah jesus

alkico:


src/clue.js
```diff
@@ -1,14 +1,146 @@
...
+    {rope: 10},
+    {knife: 8},
+    {candlestick: 2},
+    {dumbbell: 30},
+    {poison: 2},
+    {axe: 15},
+    {bat: 13},
+    {trophy: 25},
+    {pistol: 20},    
+    ];
```
Jeff-Moraes: You don't have to create an object like this actually. You can rather just write:
`const weaponsArray = [
  { name: "rope", weight: 10 },
  { name: "knife", weight: 8 },
  { name: "candlestick", weight: 2 },
  { name: "dumbbell", weight: 30 },
  { name: "poison", weight: 2 },
  { name: "axe", weight: 15 },
  { name: "bat", weight: 13 },
  { name: "trophy", weight: 25 },
  { name: "pistol", weight: 20 }
];`

src/clue.js
```diff
@@ -1,14 +1,146 @@
...
 
 // ITERATION 2
+//Step 1 = create a function that randomly selects
+// a card (ie object/element in array) from one of the collections (ie arrays)
+function selectRandom (arr) {
+    let randomIndex = Math.floor(Math.random()* arr.length)
+    let card = arr[randomIndex];
+    console.log(card);
+    return card;
+}
```
Jeff-Moraes: well done!

src/clue.js
```diff
@@ -1,14 +1,146 @@
...
+function pickMystery () {
+   
+    const mystery = {
+    suspect: selectRandom (suspectsArray),
+    room: selectRandom (roomsArray),
+    weapon: selectRandom (weaponsArray)
+  }; 
+    return mystery;
+    
+}
```
Jeff-Moraes: You don't have to create an object like this actually. You can rather just write:
const newObject = {}
and then assign the new proprietors to it.

src/clue.js
```diff
@@ -1,14 +1,146 @@
...
+
+pickMystery();
 
 // ITERATION 3
+//step 1 create function
+//step 2 function receives envelope (that contains suspct, weapon, room) object as argument
+//step 3 return message that reveals contents of envelope
+/*
+function revealMystery(obj) {
```
Jeff-Moraes: The solution to this interaction is:
`function revealMystery(mysteryEnvelope) {
  return `${mysteryEnvelope.suspect.firstName} ${mysteryEnvelope.suspect.lastName} killed Mr. Boddy using the ${mysteryEnvelope.weapon.name} in the ${mysteryEnvelope.room.name}!`;
}`

PR #1660, opened 2020-05-06
AMS WD FT MAY 2020 - Aïda Loemba

aidaloemba:
done

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,14 +1,124 @@
 // ITERATION 1
 
 // Suspects Collection
-const suspectsArray = [];
+const suspectsArray = [
+    mrGreen = {
+        firstName: 'Jacob',
+        lastName: 'Green',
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -1,14 +1,124 @@
...
+    let mystery = {
+        suspect: selectRandom(suspectsArray),
+        weapon: selectRandom(weaponsArray),
+        room: selectRandom(roomsArray)
+    };
+    return mystery
+};
 
 // ITERATION 3
+function revealMystery() {
```
mswidzinska, line 118: here you should use an argument for a function and envelope is not = pickMystery()

    return (
    envelope.suspect.firstName +
    " " +
    envelope.suspect.lastName +
    " killed Mr. Boddy using the " +
    envelope.weapon.name +
    " in the " +
    envelope.room.name +
    "!"
  );
};

PR #1718, opened 2020-06-10
[Remote-PT-062020] Nuno Pacheco

nuno-pacheco:
Without Stringify, I have to read about that.


valmgisbert:
You're getting better at cleaning your code! nice! You can refactor your first two functions, there are certain parts you're not really using or that are unnecessary. Be careful with syntax errors. Overall nice approach! Great job!

PR #1719, opened 2020-06-10
[Remote-PT-0620] Debora Crescenzo 

NubivagantYansa:
Exercise completed :)

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -137,18 +137,14 @@ const selectRandom = arr => {
 selectRandom(roomsArray)
 
 const pickMystery = () =>{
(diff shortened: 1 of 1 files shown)


valmgisbert:
I was about to tell you that you can refactor the third function but you're too fast for me. AAA THE STRINGIFY!!! there was no need to use it there (I mentioned it for the links and probably should have used another word instead of making my own verb of string lolol) but it's wonderful that you're researching the methods, I love it! Wonderful job!

NubivagantYansa:
Just learned about object destructuring and refactored my code

PR #1720, opened 2020-06-10
[REMOTE-WDPT-0602] - Sunitha Mannam

SunithaMannam:


valmgisbert:
You can refactor your 2nd iteration; nice approach though, it's very meticulous. Love it! Excellent job!

PR #1721, opened 2020-06-10
[REMOTE-WDPT-0602] – Willem Prins

willemprins93:
Did it.


valmgisbert:
`Math.floor()` for your `arr.length` is a bit redundant since it always gives you an integer. Remember to use string interpolation to be more interactive with your variables. Excellent work! Very concise and thorough.

PR #1722, opened 2020-06-10
[REMOTE-WDPT-0602] - Giovanna Izzo

Hachiko9:


valmgisbert:
`arr.length` always gives you an integer, so using `Math.floor()` is unnecessary. What lovely code! Very organised and thorough. I like that you used ternaries for an added failsafe. Great job!

PR #1735, opened 2020-06-11
[REMOTE-WDPT-0602] Cristina Matamoro

cris-developer:
src/clue.js

Not sure if Jasmine is working fine. On other code editors i checked out the answers and that meet the requirements (fi, that some functions retrieves and object,..)

Happens the same with Jasmine on the former lab. It was informing they were right, the exercises, and in some point, without changing the code, retrieves was wrong. The code works on any other editor and meets the requirements

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,14 +1,107 @@
 // ITERATION 1
 
 // Suspects Collection
-const suspectsArray = [];
-
-// Rooms Collection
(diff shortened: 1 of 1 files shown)


valmgisbert:
Nice approach! You can refactor it a bit more. Remember to use string interpolation for your final string result. Good job!

PR #1738, opened 2020-06-11
[WD PT REMOTE 062020] - Lynn W-B

lynn-w-b:
Completed all iterations but could not get last test parameter to work on Jasmine.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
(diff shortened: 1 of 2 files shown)


valmgisbert:
Check where you are invoking the function that you made for the random cards to avoid repeating yourself, and for the last test check also the parameter of your `revealMystery` function to see where it connects. Refactor your code to keep it concise and avoid putting repeating code as well. Nice approach, though!

PR #1739, opened 2020-06-11
[REMOTE-WDPT-0602020] - Rodrigo Hernández

Rodrigohernandezsan:
Exercise completed. I had difficulties in the last part (Create the mystery).

I also have some problems with (Jasmine) sometimes it doesn't work well.

Thank you

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,14 +1,110 @@
 // ITERATION 1
 
 // Suspects Collection
-const suspectsArray = [];
+const suspectsArray = [
 
+    {nickname: "mrGreen",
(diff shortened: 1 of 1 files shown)


valmgisbert:
The random selection approach is good, but your multiple function declarations + uses are completely differentiating from the instructions, which will make Jasmine not work since it won't recognise anything to test. Please check carefully your code; a lot of typos can render your methods useless. Your last function has no parameters so it won't connect to anything previous in your code when you invoke it. Also, please practice string interpolation for the returned result. Whenever you can, please redo this LAB.

PR #1740, opened 2020-06-11
[REMOTE-WDPT-0602] - Alejandro Gonzalez

alejanglez:
Jasmine is telling me this last error - Expected '"Doctor Orchid killed Mr.Boddy using the poison in the Living Room"' to equal 'Victor Plum killed Mr. Boddy using the poison in the Billiard Room!'.

I have tried stringify, parse, or both. Impossible!

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -115,10 +115,11 @@ let envelope = pickMystery();
 
 function revealMystery(obj) {
  
-  return (envelope.suspect.firstName + " " + envelope.suspect.lastName + " killed Mr.Boddy using the " 
-+ envelope.weapon.name + " in the " + envelope.room.name);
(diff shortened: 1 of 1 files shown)


valmgisbert:
Be careful with spacing in your method invocation, also clean your console.logs. Remember that you don't need to console.log an invoked function. A couple of details on the final iteration to help you why it isn't working: your parameters in the final function differentiate, check typos, and a good practice is to use string interpolation; `JSON.stringify()` doesn't really do anything here since we don't need to change the string format (when I explained the link part of the object to you all, I conjugated a word without thinking of the existing JSON method and that it could confuse you; my mistake! sorry!). Other than that, good job!

PR #1741, opened 2020-06-11
[REMOTE-WDPT-0602] - Juan José Carreño

JJexcelsior:


valmgisbert:
Wonderful use of deconstructing. Very clean code. Excellent job!

PR #1757, opened 2020-07-07
[PTWDAMS062020] Anna Dorenskaia

AnnaDore:
Hi Guido! Thank you for your help! With your help I could figure out why I couldn't reach 'firstName' from the 'suspectsArray' correctly according to the tests. With your help my object started to look like this:

const finalObject = { 
        suspect: randomSuspect, 
        weapon: randomWeapons, 
        room: randomRoom
}
and I could pass the tests when I started to reach the 'firstName' like this:
object.suspect.firstName
after all problems were solved!

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -166,10 +166,9 @@ function selectRandom (array) {
 
 
 function pickMystery() {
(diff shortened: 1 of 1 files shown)


guido-an:
Well done!

PR #1873, opened 2020-10-14
[FTAMS102020] Nina van den Braak

ninsajala:
Finished the 3 iterations


src/clue.js
```diff
@@ -1,14 +1,136 @@
 // ITERATION 1
```
LloydC, line 2: Nicely done Nina, I don't see anything wrong with this solution

PR #1879, opened 2020-10-14
[FTAMS102020] Irene Asensio

ihatepineapple:


src/clue.js
```diff
@@ -1,14 +1,195 @@
...
+        return undefined
+    } else if (array.length === 1){
+        return array[0]
+    } else if (array.length > 0) {
+        return array[Math.floor(Math.random() * array.length)]
+    }
+}
 
+//Function pick Mystery
+
```
LloydC, line 180: Good job Irene! You don't need the "object" paramater in your pickMistery function. You should remove it

src/clue.js
```diff
@@ -1,14 +1,195 @@
...
+function pickMystery(object){
+    let suspect = selectRandom(suspectsArray)
+    let room = selectRandom(roomsArray)
+    let weapon = selectRandom(weaponsArray)
+
+    return envelope = {suspect, weapon, room }
```
LloydC, line 186: There's no use declaring "envelope" here, you could just write `return {suspect, weapon, room}`

ihatepineapple: I think Jasmine wasn't passing it without this object, but my first answer was just as you say

PR #1881, opened 2020-10-14
[FTAMS102020] AlejandroPliakasBenitez 

apliakas:


src/clue.js
```diff
@@ -1,14 +1,114 @@
 // ITERATION 1
```
LloydC, line 2: Good job Alejandro! Everything looks fine in your solution

PR #1899, opened 2020-10-21
[WDFTRM-20201021] - Francisco Melo

xikz:


src/clue.js
```diff
@@ -1,14 +1,131 @@
...
+function revealMystery(obj) {
+  return (
+    obj.suspect.firstName +
+    " " +
+    obj.suspect.lastName +
+    " killed Mr. Boddy using the " +
+    obj.weapon.name +
+    " in the " +
+    obj.room.name +
+    "!"
```
filipe-freire, line 129: You could also write it like: ``${obj.suspect.firstName} ${obj.suspect.lastName} killed Mr. Boddy using the ${obj.weapon.name} in the ${obj.room.name}!`;`

filipe-freire (approved):
Nice job Francisco! Congrats! 🥇

PR #1905, opened 2020-10-21
[WDFTRM-20201021] - Gosia

GosiaMas:
Filipe explained how object is inserted to last function with Jasmine


filipe-freire (approved):
Everything is awesome Gosia! Congrats 🥇

PR #1928, opened 2020-10-21
[WDFTRM 20201021] - Monique

moniquedehaard:


src/clue.js
```diff
@@ -1,14 +1,125 @@
...
+{name: "bat" , weight: 13},
+{name: "trophy" , weight: 25},
+{name: "pistol" , weight: 20},
+
+];
 
 // ITERATION 2
+function selectRandom(array){
+    if (!array.length){
+        return undefined;
```
filipe-freire, line 94: In this case, null should be returned instead of undefined. Null is used whenever you want to intentionally declare something as nonexistent. Undefined is usually returned when there is an error in our part or something is missing unintentionally.

filipe-freire (approved):

PR #1996, opened 2021-01-13
BER-FT-Jan-2021-Patrick-Lehmann-&-Aylin-Ünal

ptrcklehmann:
Patrick Lehmann and Aylin Ünal first collab


ptrcklehmann:
It works only, but the error is only solved when the message is specified as per the test

ptrcklehmann:
We managed to fix it. The function was missing _envelope_ argument.

PR #2037, opened 2021-01-21
[BLN-PT-JAN2021] Gizem Wanka

gizeminthecity:
add cluedo solution


IH-WebDev-TA-Remote (commented):
All good here Gizem! Another lab in the bag, keep up the good work! :medal_sports:

PR #2038, opened 2021-01-21
[WDPT JAN21] Aleksandra Porada: Finished all the iterations successfully

hola-ola:


src/clue.js
```diff
@@ -1,14 +1,150 @@
...
+    } return array[Math.floor(Math.random() * array.length)];
+};
+
+// Create the mystery
+
+let envelope = {
+    suspect: "?",
+    weapon: "?",
+    room: "?"
+};
```
IH-WebDev-TA-Remote, line 137: You could also have declared the envelope variable inside the function pickMystery() since it is only re-assigned there. This way, the function pickMystery() wouldn't have the side-effect of mutating a variable outside its scope, an occurrence we should avoid!

IH-WebDev-TA-Remote (commented):
All good here Aleksandra! A job (or should I say lab!) well done once again! Cheers! :medal_sports:

src/clue.js
hola-ola: Clear! Thanks.

PR #2040, opened 2021-01-21
Finishing Clue exercise.

lucajans:


src/clue.js
```diff
@@ -1,14 +1,128 @@
...
+        name: "pistol",
+        weight: 20
+    }
+];
+
 
 // ITERATION 2
 
+function selectRandom(arr) {
+    for (element of arr) {
```
IH-WebDev-TA-Remote, line 111: With this approach, a loop wouldn't be required since we can immediately return a random index of a certain array with the line 112 :bulb:

IH-WebDev-TA-Remote (commented):
Amazing job Luca! :tada: Congratulations on another lab well done!

PR #2059, opened 2021-03-23
[FTAMS032021] Andrea Sforza

arsforza:
Cluedo assignment completed


reMRKableDev (commented):
Good stuff @arsforza 🚀 

Your implementations are all green against the Jasmine tests. Good to see you making use of arrow functions in the second iteration as well

PR #2060, opened 2021-03-23
[FTWDAMS032021] - Jamal Nasser

jamal-nasser:


src/clue.js
```diff
@@ -1,14 +1,147 @@
...
+}
+//console.log(selectRandom(roomsArray));
+/////////
+function selectRandom(weaponsArray)
+{
+  
+return weaponsArray[Math.floor(Math.random()*weaponsArray.length)];
+     
+}
+//console.log(selectRandom(weaponsArray));
```
reMRKableDev, line 111: Good implementations, however, here is a good place to utilize the **D.R.Y** principle. With **Don't Repeat Yourself**, it's better to have a singular function instead of repeating the same code over and over again.

---

Your code could be written as follows and still work well
```suggestion
function selectRandom(incomingArray){
  return incomingArray[Math.floor(Math.random() * incomingArray.length)] 
}
```

---

src/clue.js
```diff
@@ -1,14 +1,147 @@
...
+
+
+//Create the mystery
+function pickMystery() {
+  
+  let room = roomsArray[Math.floor(Math.random()* roomsArray.length)]
+  
+  let weapon = weaponsArray[Math.floor(Math.random() * weaponsArray.length)]
+  
+  let suspect = suspectsArray[Math.floor(Math.random() * suspectsArray.length)]
```
reMRKableDev, line 123: You can make use of the `selectRandom()` function above to simplify this code here as well. 

```suggestion
  let room = selectRandom(roomsArray);
  
  let weapon = selectRandom(weaponsArray);
  
  let suspect = selectRandom(suspectArray)
```

src/clue.js
```diff
@@ -1,14 +1,147 @@
 // ITERATION 1
```
reMRKableDev, line 1: **A general tip from now**: Delete the console.log from the file on your final submission. Reason -> The tests don't take into account the console.log when verifying your solutions. They only consider the function.

**The exception** _It's ok to use console.log if the purpose is to be sure that the value you are getting back is correct. Once you have confirmed this, however, you may delete it._

reMRKableDev (commented):
Nice work @jamal-nasser 🚀 

All your code is passing the Jasmine tests. You can apply some improvements in your codebase to make it more D.R.Y. 

Check out the comments and apply the suggested changes.

reMRKableDev (commented):
Good work implementing the recommended changes @jamal-nasser ✅

PR #2061, opened 2021-03-23
[ftwebdevams032021] jeroen

jeroenvanbart:


src/test.js
```diff
@@ -0,0 +1,19 @@
+const roomsArray = [
```
reMRKableDev, line 1: You can delete this file when you turn in your final solution

src/clue.js
```diff
@@ -1,14 +1,120 @@
 // ITERATION 1
```
reMRKableDev, line 1: **A general tip from now**: Remove the `console.log` on your final submission. Reason -> The tests don't take into account the `console.log` when verifying your solutions. They only consider the function 

**The exception**: _It's ok to use `console.log` if the purpose is to be sure that the value you are getting back is correct. Once you have confirmed this, however, you may delete it_

src/clue.js
```diff
@@ -1,14 +1,120 @@
...
+
+const selectRandom = (array) => {
+    let randomObject = array[Math.floor(Math.random() * array.length)];
+    return (randomObject);
+};
+
+const pickMystery = () => {
+    let mystery={};
+
+mystery = {suspect:selectRandom(suspectsArray), weapon:selectRandom(weaponsArray), room:selectRandom(roomsArray)};
```
reMRKableDev, line 106: This could've been formatted better for readability. Feel free to install the extension [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) on your code editor. It helps do these things automatically.

reMRKableDev (commented):
Super work @jeroenvanbart 🚀 

There aren't any failing aspects of your solution. This is very good. 
I have left some general tips for you to consider for your future submissions

PR #2085, opened 2021-03-24
[FTAMS032021] SilviaRodriguez

silrodcar:


src/clue.js
```diff
@@ -1,14 +1,142 @@
...
+color: "green"
+},
+
+{
+nickname: "drOrchid",
+firstName: "Doctor",
+lastName: "Orchid",
+occupation: "Scientist",
+age: "26",
+description: "PhD in plant toxicology. Adopted daughter of Mr. Boddy"
```
reMRKableDev, line 24: There needs to be a comma here. Otherwise, the whole file won't be read properly by Jasmine

```suggestion
description: "PhD in plant toxicology. Adopted daughter of Mr. Boddy",
```

src/clue.js
```diff
@@ -1,14 +1,142 @@
...
+
+
 
 // ITERATION 2
 
-// ITERATION 3
+//Random selector:
+
+function selectRandom (arr) {
+    let = randomCard = arr[Math.floor(Math.random()*arr.length)];
```
reMRKableDev, line 122: The equal sign isn't necessary here

```suggestion
    let randomCard = arr[Math.floor(Math.random()*arr.length)];
```

src/clue.js
```diff
@@ -1,14 +1,142 @@
...
+
+{
+nickname: "missScarlet",
+firstName: "Kasandra",
+lastName: "Scarlet",
+occupation: "Actor",
+age: "31",
+description: "She is an A-list movie star with a dark past",
+image: "https://www.radiotimes.com/uploads/images/Original/111967.jpg",
+color: red
```
reMRKableDev, line 48: So the main cause for `pickMystery` failing the tests is this here. The value `red` needed to be made into a string `"red"`
```suggestion
color: "red"
```

reMRKableDev (commented):
Nice @silrodcar 🚀 

I left some comments that should make sure your Jasmine tests are green. Like I mentioned in our call, you are implementing everything correctly as far as the function. The problem lied in the `suspectsArray`. I added a comment on how you can fix this issue

PR #2086, opened 2021-03-25
[FTAMS032021] Igor Godinho Figueiredo

Igor-GF:
I could reach to end but Jasmine shows pending steps that I couldn't fix.

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -126,22 +126,19 @@ selectRandom = (pickRandom) => {
 pickMystery = () => {
   let foundOutMystery = {};
 
-  let toGetFullName = selectRandom(suspectsArray);
-  //random first and last names asigned to variable foundMystery
-  foundOutMystery.suspect = `${toGetFullName.firstName} ${toGetFullName.lastName}`
+  foundOutMystery.suspect = selectRandom(suspectsArray);
   // random weapon asigned to variable foundMystery
-  foundOutMystery.weapon = `${selectRandom(weaponsArray).name}`
+  foundOutMystery.weapon = selectRandom(weaponsArray);
   //random room asigned to variable foundMystery
-  foundOutMystery.room = `${selectRandom(roomsArray).name}`
+  foundOutMystery.room = selectRandom(roomsArray);
 
   return foundOutMystery;
 }
 
 // ITERATION 3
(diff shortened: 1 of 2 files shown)


src/clue.js
```diff
@@ -1,14 +1,147 @@
...
+pickMystery = () => {
+  let foundOutMystery = {};
+
+  let toGetFullName = selectRandom(suspectsArray);
+  //random first and last names asigned to variable foundMystery
+  foundOutMystery.suspect = `${toGetFullName.firstName} ${toGetFullName.lastName}`
+  // random weapon asigned to variable foundMystery
+  foundOutMystery.weapon = `${selectRandom(weaponsArray).name}`
+  //random room asigned to variable foundMystery
+  foundOutMystery.room = `${selectRandom(roomsArray).name}`
```
reMRKableDev, line 135: A good thought process here, however, you could simply call the `selectRandom()` method directly to the property in `foundOutMystery` object. See suggestion 

```suggestion
  foundOutMystery.suspect = selectRandom(suspectsArray);
  foundOutMystery.weapon = selectRandom(weaponsArray);
  foundOutMystery.room = selectRandom(roomsArray);
```

src/clue.js
```diff
@@ -1,14 +1,147 @@
...
+
+  return foundOutMystery;
+}
 
 // ITERATION 3
+revealMystery = (envelope) => {
+  return `${envelope.suspect} killed Mr. Boddy using the ${envelope.weapon} in the ${envelope.room}`;
```
reMRKableDev, line 142: Remember that `envelop` will be the full object that is returned by `pickMystery`. So in order to get the correct data, you need to access correct keys from `suspect`, `weapon`, and `room`. 

```suggestion
  return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`
```

reMRKableDev (commented):
Nice work @Igor-GF 🚀 

`pickMystery` and `revealMystery` need some corrections in order to pass all the Jasmine tests. I've left some comments you can check out about that

reMRKableDev (commented):
🚀 🚀 🚀

PR #2144, opened 2021-05-25
[WDPT-Berlin-May2021] Johannes Polte 

jmpberlin:
-created all objects
-created all functions

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,14 +1,154 @@
 // ITERATION 1
 
 // Suspects Collection
-const suspectsArray = [];
-
-// Rooms Collection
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -1,14 +1,154 @@
...
 
+
+    function selectRandom(arr){
+
+        return arr[Math.floor(Math.random()*arr.length)];
+    };
+
+    
+    function pickMystery (){
+        let mystery=new Object();
```
numbing, line 143: good job, just small thing: i think its better just simply write `let mystery = {}`

PR #2210, opened 2021-06-06
Lily - WDFT - RMT - MAY2021 - Clue solved

liliiakhr:


TA-Remote:
Great Job Lily! 💯 You did all the mandatory and also the bonus. Keep the hard work! 🎖️ Making the remove buttons work for the ones you created wasn't that hard. I know you can! 👍 Keep the hard work! You are doing great!

PR #2235, opened 2021-07-14
[WD PTRM - 0721] Pablo Alonso

pablochux:


IH-WebDev-TA-Remote:
Great work Pablo! 💪🏻Yet another lab in the bag. 😎 Here are some comments:
-> On the `pickMystery()` function, you are initializing the obj as an array, but the lab expects the returned value to be an object with the properties "suspect", "weapon" and "room". Although your solution works, it would be drier if you returned and object with those properties already, such as:
`
return {
  }
`
-> On the last iteration, you could've also used the template literal syntax (backticks), to make the returned value a bit more readable! (e.g: `return `${mystery.suspect.firstName} ${mystery.suspect.lastName} killed Mr. Boddy using the ${mystery.weapon.name} in the ${mystery.room.name}!`;`

PR #2237, opened 2021-07-15
Cluedo - Noemi Radvan WDPT JUL 2021

codingfannn:
Please check and leave comments if  not correct!

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,22 +2,211 @@
 
 // Suspects Array
 
-const suspectsArray = [];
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Great job so far Noemi! 🎉Regarding the code:
-> You can rename "suspects" on the `pickMystery()` function to "suspect", since it's the property that the test expects in the final object!
-> Only Iteration 3 to go! 👀 Hope you can give it another try! 💪🏻

PR #2250, opened 2021-07-23
[WD-PT-Remote] Juan Carlos  

juaneugenio:


IH-WebDev-TA-Remote:
Awesome job Juan! Another lab completed! :trophy: Here are some comments regarding the code:
-> In iteration 2 you don't need to loop through the array passed as the argument. Instead you can just keep the return statement you have there.
Apart from that, excellent work! Keep it up :muscle:

PR #2334, opened 2021-09-21
WDPT-REMOTE-BRIGIHORVATH

brigihorvath:


src/clue.js
```diff
@@ -2,29 +2,128 @@
...
+  };
 
-function revealMystery() {}
+  return mystery;
+}
 
+// ITERATION 3
 
+function revealMystery(envelope) {
+  return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`;
```
tomasyaya, line 125: we are going to see destructuring later on but know that you can do something like

```
function revealMystery({suspect, room, weapon}){
}
```

tomasyaya (commented):
Great work @brigihorvath

PR #2337, opened 2021-09-21
[REMOTE-PT-092021]Philipp Kastl

ignazka:


src/clue.js
```diff
@@ -2,29 +2,179 @@
...
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
+  return arr[Math.floor(Math.random() * arr.length)];
+}
 
-function pickMystery() {}
+function pickMystery() {
```
tomasyaya, line 164:
```
const randomStack = {
   suspect: selectRandom(suspectArray),
   weapon: selectRandom(waponArray)
   
   ...
}
```

src/clue.js
```diff
@@ -2,29 +2,179 @@
...
+  return randomStack;
+}
 
 // ITERATION 3
 
-function revealMystery() {}
-
-
+function revealMystery(obj) {
+  return `${obj.suspect.firstName} ${obj.suspect.lastName} killed Mr. Boddy using the ${obj.weapon.name} in the ${obj.room.name}!`;
```
tomasyaya, line 176: we are going to see destructuring later on but know that you can do something like

```
function revealMystery({suspect, room, weapon}){
}
```

tomasyaya (commented):
Great work!
@ignazka

PR #2346, opened 2021-09-23
WDPT-2021-REMOTE-Lilli-Reilly

l-reilly:
It works but it is not passing all the tests. I don't know why it is saying it is not returning a string, and I'm unsure why the pickMystery function isn't passing.

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -122,7 +122,8 @@ function pickMystery() {
 // ITERATION 3
 let envelope = pickMystery(suspectsArray, roomsArray, weaponsArray);
 function revealMystery(envelope) {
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,27 +2,129 @@
...
+  let selectWeapons = selectRandom(weaponsArray);
+  return [selectSuspect, selectRoom, selectWeapons]
+}
 
 
 // ITERATION 3
-
-function revealMystery() {}
+let envelope = pickMystery(suspectsArray, roomsArray, weaponsArray);
+function revealMystery(envelope) {
```
tomasyaya, line 124: we are going to see destructuring later on but know that you can do something like

```
function revealMystery([suspect, room, weapon]){
}
```

tomasyaya (commented):
Very nice! @l-reilly

PR #2348, opened 2021-09-23
WD PT SET 2021 JOANA BRITO

jobritomar:


src/clue.js
```diff
@@ -2,27 +2,176 @@
...
 
 
 // ITERATION 2
 
-function selectRandom() {}
 
-function pickMystery() {}
+
+
+function selectRandom(playedCard) {
```
tomasyaya, line 132: So first thing is try to indent properly the function body so that is easier to read.

```
function selectRandom(playedCard){
// this is a bit redundant because if you do something like [(emptyArray)][1] it will give undefined
 if(playedCard.length === 0) {
   return undefined
 }
 
 // You missed Math.floor that will round down your number
 // For instance if Math.random gives 3.3 then Math.floor will return 3
 const randomIndex = Math.floor(Math.random() * playCard.length)
 
 const card = playedCard[randomIndex]
 return cad
}
```

tomasyaya (commented):
Great work working on this @jobritomar

PR #2435, opened 2021-11-09
Diogo & Rach PT Lisbon

diogoascarneiro:
Me and Rach both did it in our own computers because of different os' / keyboards, but we got to the answers together :)

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,29 +2,130 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: 'Jacob',
+    lastName: 'Green',
+    occupation: 'Entrepreneur',
+    age: 45,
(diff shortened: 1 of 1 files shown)


carlarsmendes:
Hi guys! 
Thanks for the clarification Diogo!
So I found the issue on your `randomIntFromInterval`  function:
With this `return Math.floor(Math.random() * (max - min) + min)` it should work, next time test your code with (for eg.) a small array of two elements and make sure you are always getting only those 2 - Sometimes I was getting undefined, so your min/max were out of bounds for your array.
I'm also missing something like this `if(arr.length===1) return arr[0];`  in `selectRandom` on this pull request, but I had the impression all your remaining tests were green, so maybe you haven't committed this part.
Good job, your code looks straightforward and simple. Hope you enjoyed this one :)

PR #2443, opened 2021-11-10
PT Lisbon André & Corentin

CorentinAesch:


carlarsmendes:
Hi guys! Great job on finishing the lab.
A couple of small notes:
Iteration 2-> Look out for your variable name! Maybe instead of "pick", you could have `randomIndex` ?
Iteration 3-> It wasn't passing the test because your function isn't working with its own arguments. The function revealMistery needs an input `envelope`
Then you would need to replace pickMistery() with envelope, like this:
```
  let reveal = `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${usedWeapon = envelope.weapon.name} in the ${envelope.room.name}!`;
  return reveal;
}
```
And like this your tests will pass :) 
Then to call your function, you simply need to write:
revealMystery(pickMistery());

Another tip -> Your pickMistery function is all correct, but you could even save up a bit of code and write it like this:

```
function pickMystery() {
  return mystery = {
  };
}
```
Great work!

PR #2493, opened 2022-01-12
Nico Tollenaar - completed

NicoTollenaar:


src/clue.js
```diff
@@ -2,27 +2,151 @@
...
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(array) {
+  let randomIndex = Math.floor(Math.random()*(array.length));
+  let randomCard = array[randomIndex];
+  console.log("randomIndex: ", randomIndex);
+  return randomCard;
+}
```
TA-Remote, line 131: Great, I have nothing to say except that I love how you structure the code. I mean, instead of returning array[Math.floor(Math.random() * array.length)], you store the result in a variable and it's much easier to read. It's just little things here, but this is super helpful especially in later projects and collaboration. 
However, there are also enough people who think: the shorter the code, the better. That is probably a matter of taste.

PR #2496, opened 2022-01-12
[WDFT-Jan 22] Day 3 Lab#2 - Doan Tran

Dorrrn:
Changes in clue.js

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,29 +2,130 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: 'Jacob',
+    lastName: 'Green',
+    occupation: 'Entrepreneur',
+    age: 45,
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,29 +2,130 @@
...
-function selectRandom() {}
-
-function pickMystery() {}
+function selectRandom(arr) {
+  if (!arr || arr.length === 0) {
+    return undefined;
+  } else if (arr.length === 1) {
+    return arr[0];
+  } else {
+    const random = Math.floor(Math.random()) * arr.length;
```
TA-Remote, line 109: On line 109 you out brackets in the wrong place -> so test is not passing. 

First you want to multiply Math.random() with length, and only then floor it.

Dorrrn: thanks for reviewing :)

PR #2499, opened 2022-01-12
[Remote-FT-012022] Zena Bos

zenabos:


src/clue.js
```diff
@@ -2,29 +2,185 @@
...
-function revealMystery() {}
-
+function revealMystery(envelope) {
+  let firstName = envelope.suspect.firstName;
+  let lastName = envelope.suspect.lastName;
+  let weapon = envelope.weapon.name;
+  let room = envelope.room.name;
 
+  return `${firstName} ${lastName} killed Mr. Boddy using the ${weapon} in the ${room}!`;
+}
```
TA-Remote, line 183: Really good! And I like that you store everything in a variable, it's much more readable

PR #2504, opened 2022-01-12
WD Remote Jan2022 - Ana Pavao

pavaoana:
I wasn't able to find a solution to the last iteration

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,29 +2,125 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: 'Jacob', // mrGreen
+    lastName: 'Green',
+    occupation: 'Entrepreneur',
+    age: 45,
+    description: 'He has a lot of connections',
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,29 +2,125 @@
...
 // ITERATION 3
 
-function revealMystery() {}
-
-
+function revealMystery({}) {
+  //I really don't know what to write here (it's probably very simple...)
+  //I also don't know what's missing in the line below
+  return `${suspectsArray.firstName} ${suspectsArray.lastName} killed Mr. Body using the ${weaponsArray.name} in the ${roomsArray.name}!`;
+}
```
TA-Remote, line 123: Almost!

- your function received an object -> so you want to pass any name `function revealMystery(myObject)`
- then before you start any logic, I advise console.log information that you receive `console.log(myObject)`, so you know the structure of your data
- Now I believe you should be able to fix line122 accordingly. Let me know if still not clear! Karina

PR #2638, opened 2022-03-30
[FTRMT032022] Oscar Gonzalez

oscargonzalezdev:
- [x] solve the lab-javascript-clue

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,28 +2,184 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: 'Jacob',
+    lastName: 'Green',
+    occupation: 'Entrepreneur',
+    age: 45,
+    description: 'He has a lot of connections',
+    image: 'https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg',
+    color: 'green'
+  },
+  {
+    firstName: 'Doctor',
+    lastName: 'Orchid',
+    occupation: 'Scientist',
+    age: 26,
+    description: 'PhD in plant toxicology. Adopted daughter of Mr. Boddy',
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,28 +2,184 @@
...
+let newObject = {
+  suspect: '',
+  weapon: '',
+  room: ''
+};
+newObject.suspect = randomSuspect;
+newObject.weapon = randomWeapon;
+newObject.room = randomRoom;
+return newObject;
+}
```
TA-Remote, line 175: Code could be refactored here.

First step, either you simply declare a variable newObject as empty object like this` let newObject = {}`, since you are adding properties in lines 171-173, or youcan build your object straight away like this:

```
let newObject = {
suspect: randomSuspect,
weapon: randomWeapon,
room: randomRoom,
};
```

Next step to refactor could be to use function directly to set the value, like this:
```
let newObject = {
...
```

However only use suggested syntax if you fully understand :)

src/clue.js
```diff
@@ -2,28 +2,184 @@
...
 
 // ITERATION 3
 
-function revealMystery() {}
-
+function revealMystery(newObject) {
+  let msg = `${newObject.suspect.firstName} ${newObject.suspect.lastName} killed Mr. Boddy using the ${newObject.weapon.name} in the ${newObject.room.name}!`;
+  return msg;
+}
```
TA-Remote, line 182: If you are not going to use the variable, no need to declare it. So you can return result directly:

```
function revealMystery(newObject) {
  return `${newObject.suspect.firstName} ${newObject.suspect.lastName} killed Mr. Boddy using the ${newObject.weapon.name} in the ${newObject.room.name}!`;
}
```

TA-Remote (commented):
Keep it up Oscar!

PR #2640, opened 2022-03-30
WDFT - MARC2022 - RUBEN POELEN

coderKron:
Finished it.

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,30 +1,157 @@
+/*
+
+Clue has three different card types: suspects, rooms, and weapons.
+
+Let's create an array for every card type. These arrays should be named suspectsArray, weaponsArray, roomsArray. Each array will contain objects, where each object represents one card.
+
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -1,30 +1,157 @@
...
-
+function revealMystery(object) {
+  let revealMessage;
+  for (const key in object) {
+    console.log(object['suspect']);
+    // console.log(object[key].firstName, object[key].lastName, object[key].name);
+    revealMessage = `${object['suspect'].firstName} ${object['suspect'].lastName} killed Mr. Boddy using the ${object['weapon'].name} in the ${object['room'].name}!`;
+  }
+  return revealMessage;
+}
```
TA-Remote, line 151: no need to loop here, it works without it

TA-Remote (commented):
Great work! Firstly please remove all unused code ;) Like console.log and calling functions. It's a very good habit to have to write clean code! 

I suggest reviewing alternative solution -  we'll be uploading later on student portal. Your code could be more simple :)

PR #2644, opened 2022-03-30
[Remote-FT-032022] Thibaut LARTIGUE

thibautDevWorld:


src/clue.js
```diff
@@ -2,27 +2,137 @@
...
 
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
+  let randomElement = 
+ arr[Math.floor(Math.random() * arr.length)]
+ return randomElement
+}
```
TA-Remote, line 111: here for even shorter syntax, since you are not going to use this variable you can return result directly. No need to save it beforehand :) like this:

```
function selectRandom(arr) {
return  arr[Math.floor(Math.random() * arr.length)]
}
```

TA-Remote (commented):
Great work, nothing to add! Keep it up

PR #2645, opened 2022-03-30
WDFT Mar2022 Pauline Schnoebelen

Paulinettt:


src/clue.js
```diff
@@ -2,27 +2,128 @@
...
-function selectRandom() {}
+function selectRandom(array) {
+  if (array.length === 0) {
+    return undefined;
+  }
+  let randomIndex = Math.floor(Math.random() * array.length);
+  
+  let randomElement = array[randomIndex];
+  return randomElement;
+}
```
TA-Remote, line 108: Great work Pauline! Nothing to add really :)

A suggestion to shorten syntax - but only if you understand it! Ask if not clear :)
Instead of lines 104-107
```
```

K.

PR #2646, opened 2022-03-30
WD FT 22March Diana Moisin

didicozaur:


TA-Remote:
Great job, Diana!!

src/clue.js
```diff
@@ -2,29 +2,181 @@
...
+  };
+}
 
 // ITERATION 3
+// const envelope = pickMystery()
 
-function revealMystery() {}
-
-
+function revealMystery(envelope = pickMystery()) {
```
TA-Remote, line 177: no need to say envelope = pickMystery(), because it's in the test, they pass it there! (just envelope would work)
But that definitely shows that you get it! Nice :)

PR #2647, opened 2022-03-30
WDFT 2022March Valentin Paul

Valentin-Paul:
Didn't get the revealMystery function to work..

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,29 +1,190 @@
 // ITERATION 1
 
 // Suspects Array
+const mrGreen = {
+firstName: 'Jacob',
+lastName: 'Green',
+occupation: 'Entrepreneur',
+age: 45,
+description: 'He has a lot of connections',
+image: 'https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg',
+color: 'green'
+}
 
-const suspectsArray = [];
+const drOrchid = {
+firstName: 'Doctor',
+lastName: 'Orchid',
+occupation: 'Scientist',
+age: 26,
+description: 'PhD in plant toxicology. Adopted daughter of Mr. Boddy',
+image: 'http://www.radiotimes.com/uploads/images/Original/111967.jpg',
+color: 'white'
+}
+
+const profPlum = {
+firstName: 'Victor',
+lastName: 'Plum',
+occupation: 'Designer',
+age: 22,
+description: 'Billionaire video game designer',
+image: 'https://66.media.tumblr.com/ee7155882178f73b3781603f0908617c/tumblr_phhxc7EhPJ1w5fh03_540.jpg',
+color: 'purple'
+}
+
+const missScarlet = {
+firstName: 'Kasandra',
+lastName: 'Scarlet',
+occupation: 'Actor',
+age: 31,
+description: 'She is an A-list movie star with a dark past',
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -1,29 +1,190 @@
...
+age: 62,
+description: 'He is a former football player who tries to get by on his former glory',
+image: 'https://static.independent.co.uk/s3fs-public/thumbnails/image/2016/07/04/08/unspecified-3.jpg',
+color: 'yellow'
+}
+
 
-const suspectsArray = [];
+
+const suspectsArray = [
```
TA-Remote, line 67: You are creating your suspectsArray here, so no need to create individual suspects above

src/clue.js
```diff
@@ -1,29 +1,190 @@
...
 
-// ITERATION 3
+function pickMystery() {
+const randomMystery ={};
+const randomSuspect = suspectsArray[Math.floor(Math.random() * suspectsArray.length)];
+randomMystery.suspect = randomSuspect;
+const randomRoom = roomsArray[Math.floor(Math.random() * roomsArray.length)];
+randomMystery.room = randomRoom;
+const randomWeapon = weaponsArray[Math.floor(Math.random() * weaponsArray.length)];
+randomMystery.weapon = randomWeapon;
```
TA-Remote, line 174: You already created a fucntion to get random element from array.So you can use it here! the power of functions :)

for example  
`const randomSuspect = selectRandom(suspectsArray)`
`const randomSuspect = selectRandom(roomsArray)`

src/clue.js
```diff
@@ -1,29 +1,190 @@
...
+    color: 'green'
+  }
+];
+
+// [mrGreen, drOrchid, profPlum, missScarlet, mrsPeacock, mrMustard];
 
 // Rooms Array
 
-const roomsArray = [];
+const roomsArray = ['Dining Room', 'Conservatory', 'Kitchen', 'Study', 'Library', 'Billiard Room', 'Lounge', 'Ballroom', 'Hall', 'Spa', 'Living Room', 'Observatory', 'Theater', 'Guest House', 'Patio'];
```
TA-Remote, line 133: task says that each room should have a name - so roomsArray should be array of objects

src/clue.js
```diff
@@ -1,29 +1,190 @@
...
 
+// ITERATION 3
+
+function revealMystery(randomMystery) {
+pickMystery();
+const solution ='';
+// solution += randomMystery.suspect.firstName + ' ' + randomMystery.suspect.lastName + ' killed Mr.Boddy using the ' +randomMystery.weapon.weapon +' in the ' + randomMystery.room;
+return solution;
+}
+revealMystery();
```
TA-Remote, line 187: Lines 182 and 187 are unnecessary..

To help you figure out why it is not desired result, console.log `randomMystery`. So for example you can see below, you should access weapon by saying `randomMystery.weapon.name`:

```
{
  suspect: { firstName: 'aa', lastName: 'abc' },
  weapon: { name: 'abd' },
  room: { name: 'abb' }
}
```

And why test is not passing at the ned is the silliest thing! In the task , string to match has white space between` Mr.` and `Boddy`
`<FIRST NAME> <LAST NAME> killed Mr. Boddy using the <WEAPON> in the <ROOM>!`

PR #2648, opened 2022-03-30
WDFT Remote Lab Javascript Clue - Ha Thu TRAN

HaThu6989:


src/clue.js
```diff
@@ -2,28 +2,208 @@
...
+]
+
+// ITERATION 2
+
+function selectRandom(someArray) {
+  let random = someArray[Math.floor(Math.random() * someArray.length)]
+  return random
+}
+let suspectRandom = selectRandom(suspectsArray)
+let weaponRandom = selectRandom(weaponsArray)
```
TA-Remote, line 181: Hi Thu! All in all great job, but I would do this inside of the pickMystery function

TA-Remote: like this:
```
function pickMystery() {
  let mystery = {
  }
  return mystery
}
```

PR #2652, opened 2022-03-31
passes 22/30 tests

marija-strah:


src/clue.js
```diff
@@ -2,29 +2,75 @@
...
+];
 
 
 // ITERATION 2
 
-function selectRandom() {}
-
-function pickMystery() {}
+function selectRandom(suspectsArray) {
+  let murderer = suspectsArray[Math.round(Math.random()*(suspectsArray.length-1))]
```
TA-Remote, line 52: Here's is generic syntax to get smth random, always `Math.floor ` it, we don't want to round up.. And multiply by length 

`suspectsArray[Math.floor(Math.random()*(suspectsArray.length))]`

src/clue.js
```diff
@@ -2,29 +2,75 @@
...
+  //return randomizer;
+  return murderer;
+}
 
+function pickMystery() {
+let mysteryEnvelope =[];
+  mysteryEnvelope.push(murderer);
+    //console.log(mysteryEnvelope);
+    return mysteryEnvelope;
+}
```
TA-Remote, line 67: To pick mystery you want to use your function above, ans pick random suspect from suspect array etc

TA-Remote (commented):
Good start, keep working :)

If you have doubts please ask one of us to explain the iterations! Or give it another go on the weekend :)

K.

PR #2663, opened 2022-04-13
[FTLIS042022] Chloé Faurie Cluedo done

chloe4E:


src/clue.js
```diff
@@ -2,27 +2,130 @@
...
 
-function pickMystery() {}
+      // get random item
+      const item = arr[randomIndex];
+  
+      return item;
+}
+
+function pickMystery() {
+  let myMytsteryObj = {};
```
webLisbonTA, line 111: Code optimization hint: you don't need to declare an empty object before, you could create the object already with the key value pairs, as you already have access to the information

webLisbonTA (commented):
Congrats, Chloé! Keep the good work!

PR #2671, opened 2022-04-13
[FTLIS042022] matildeCosta

matildedacosta:


src/clue.js
```diff
@@ -1,29 +1,163 @@
...
+/* Declare a function named selectRandom to randomly select one element from a card stack. 
+The function should expect an array as an argument, and should return a random element from the array. */
+
+function selectRandom(array) {
+  let randomArray= [];
+  
+  randomArray.push(array[Math.floor( Math.random() * array.length)]);
+
+  return randomArray[0];
+}
```
webLisbonTA, line 122: This works, but actually you don't need to create an array and return the element, you could return it directly. Like this:

```
function selectRandom(arr) {
  const randomCard = arr[Math.floor(Math.random() * arr.length)];
  return randomCard;
}
```

webLisbonTA (commented):
Well done, Matilde! Keep up the good work!

PR #2678, opened 2022-04-13
[FTWD042022] Catarina Salvado

catarinapsalvado:


src/clue.js
```diff
@@ -2,29 +2,138 @@
...
+}
 
-function pickMystery() {}
+console.log(selectRandom(weaponsArray));
 
+function pickMystery() {
+  let randomSuspect = selectRandom(suspectsArray);
+  let randomWeapons = selectRandom(weaponsArray);
+  let randomRooms = selectRandom(roomsArray);
+  let fullMystery = {};
```
webLisbonTA, line 120: Code optimization hint: Instead of declaring an object, you could create the object directly with the key-value pairs.

webLisbonTA (commented):
Congrats Catarina! Keep the good work!

PR #2683, opened 2022-04-13
[FTLIS0422] Duarte Alves

drttyy:


src/clue.js
```diff
@@ -2,29 +2,130 @@
...
+  { name: 'trophy', weight: 25 },
+  { name: 'pistol', weight: 20 }
+];
 
 // ITERATION 2
 
-function selectRandom() {}
-
-function pickMystery() {}
+function selectRandom(arr) {
```
webLisbonTA, line 101: This works, but actually you don't need to create an array and return the element, you could return it directly. Like this:
```
function selectRandom(arr) {
  const randomCard = arr[Math.floor(Math.random() * arr.length)];
  return randomCard;
}
```

webLisbonTA (commented):
Well done, Duarte! Keep the good work!

PR #2684, opened 2022-04-13
done. please clarify why the last test failed. thanks

nataliabuzila:
image

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,30 +1,126 @@
+
 // ITERATION 1
 
 // Suspects Array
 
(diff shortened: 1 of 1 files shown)


ta-web-mexpt2:
function revealMystery(mistery) {
  return `${mistery.suspect.firstName} ${mistery.suspect.lastName} killed Mr. Boddy using the ${mistery.weapon.name} in the ${mistery.room.name}!`;
}

just have to receive an object with the mystery try this sintax with your last function

PR #2716, opened 2022-05-24
done

KiraTeresa:
Finished the lab passing all the tests.

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,29 +2,133 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: 'Jacob',
+    lastName: 'Green',
+    occupation: 'Entrepreneur',
+    age: 45,
+    description: 'He has a lot of connections',
+    image:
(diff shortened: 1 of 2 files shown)


src/clue.js
```diff
@@ -2,29 +2,133 @@
...
-
-function pickMystery() {}
+function selectRandom(someArr) {
+  let randomIndex = Math.floor(Math.random() * someArr.length);
+  let randomElement = someArr[randomIndex];
+  return randomElement;
+}
 
+function pickMystery() {
+  let mystery = {};
```
filipe-freire, line 115: could've also declare the mystery object like this:

```
const mystery = {
}
```

filipe-freire (commented):
Great job Kira! 🏆  Another lab in the bag! 😄 🎉 

Quick note: 

-> I noticed you use `let` for most variable declarations. While that is totally valid & correct, it's preferable to use `const` whenever possible since it does not allow for variable reassignment & indicates the developer that its value shouldn't change later in the code. ✌🏻

PR #2723, opened 2022-05-26
mystery solved

poncho678:


src/clue.js
```diff
@@ -1,30 +1,128 @@
...
+function selectRandom(arr) {
+  if (!arr.length) {
+    return undefined;
+  }
+  const randomNumber = Math.floor(Math.random() * arr.length);
+  return arr[randomNumber];
+}
 
+function pickMystery() {
+  const mysteryObject = {};
```
filipe-freire, line 109: Could've also declared the mystery object like this:

```
const mystery = {
}
```

src/clue.js
```diff
@@ -1,30 +1,128 @@
...
 
-function revealMystery() {}
-
-
+function revealMystery(cards) {
+  const {
+    suspect: { firstName, lastName },
+    weapon: { name: weapon },
+    room: { name: room }
+  } = cards;
```
filipe-freire, line 122: Uh, advanced destructuring right here! 👀 🥇

src/clue.js
poncho678: Ok cool, this is a lot shorter. Thank you

src/clue.js
filipe-freire: Definitely will be covered! 💪🏻

PR #2837, opened 2022-07-28
"Solved lab"

herrFernandez:
I din't pass the last test, but I don't know why. I tested the last function with console.log and it works well. Maybe some typo? Well, at least better thatn last lab :-D

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -204,5 +204,12 @@ function pickMystery() {
 
 function revealMystery() {
 
-    return `${pickMystery().suspect.firstName} ${pickMystery().suspect.lastName} killed Mr. Boddy using the ${pickMystery().weapon.name} in the ${pickMystery().room.name}!`;
+    const mystery = pickMistery();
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,25 +2,214 @@
...
+        room: selectRandom(roomsArray)
+    };
+
+}
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery() {
+
+    const mystery = pickMistery();
```
trebeljahr, line 207: the reason why the test is not passing is that revealMystery shouldn't create a new mystery but get an already existing mysteryObject as input. 

```javascript
function revealMystery(mystery) {
      return `${mystery.firstName} ${mystery.lastName} killed Mr. Boddy using the ${mystery.weapon} in the ${mystery.place}!`;
} 
``` 
should do the trick

trebeljahr (commented):
good job, slight misunderstanding of what the test/exercise wanted from you in the last iteration but nothing bad. added a comment explaining it.

PR #2840, opened 2022-07-30
Update clue.js

JoaoSerafim3001:


trebeljahr (commented):
you should really have another look at this lab. The rooms array as well as the persons and weapons array shouldn't be empty but filled with lots of objects - the data for these objects can be found in the DATA.md file in the project files. 

pickMystery misses completely, just try to pick a random thing from each of the arrays, (weapons, subjects, rooms) and put them as values for a mystery object which you return. 
```javascript
return {
  weapon: pickRandom(weapons),
  suspect: pickRandom(suspects),
  rooms: pickRandom(rooms)
}
``` 

remember how we picked random values from an array. combination of Math.random, Math.floor and array.length and some multiplication *

revealMystery is supposed to get an input, which is a mysteryObject - from which you would pick the correct values by dot notation or bracket notation. whichever you prefer really. 

something like mystery.subject.firstName etc.

PR #3096, opened 2023-01-18
[LIS-FT-01-23] Duarte França

Duartefranca:


src/clue.js
```diff
@@ -2,25 +2,136 @@
...
+];
 
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
+    
+    if (arr.length === 0)
+       return undefined;
```
lzaquine, line 106: Since we know this array will not return undefined, and later you return `arr[pickelement]`, you don't need these if statements.

lzaquine (commented):
Great job on the lab Duarte!

PR #3098, opened 2023-01-18
[LISFT0123] João Raposo

JRaposo22:


src/clue.js
```diff
@@ -2,25 +2,189 @@
...
+        name: 'pistol',
+        weight: 20  
+    }
+];
 
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
```
joaopdg, line 162: There's no need for the `conditions` here, because if the `array` is empty it will already return `undefined`:

```
function selectRandom(arr) {
  let randomCard = Math.floor(Math.random() * arr.length);
  return arr[randomCard];
}
```

src/clue.js
```diff
@@ -2,25 +2,189 @@
...
 
-function pickMystery() {}
+    if (arr.length === 0) return;
+    let randomCard = Math.floor((Math.random() * arr.length));
+    return arr[randomCard];
+
+}
+
+function pickMystery() {
```
joaopdg, line 170: Here you can call the `selectRandom()` function straight away as the `value` of the `keys`:

```
  function pickMystery() {
    return {
    };
  }
```

joaopdg (commented):
Great job on the lab João!

PR #3100, opened 2023-01-18
[LISFT0123] Wagner Lira

waggnerl:


src/clue.js
```diff
@@ -2,25 +2,123 @@
...
-
-function selectRandom() {}
-
-function pickMystery() {}
-
+function selectRandom(arrayData) {
+  let positionInArray = Math.floor(Math.random() * arrayData.length);
+  return arrayData[positionInArray];
+}
+function pickMystery() {
```
joaopdg, line 106: In the previous iteration you created a `function` that already returns you a random `element` of an `array` so you can use it inside this new `function` and in this way you keep it `DRY`:

```
function pickMystery() {
  return {
  };
}
```

joaopdg (commented):
Great job on the lab Wagner!

PR #3104, opened 2023-01-18
[LISFT0123] Tomás B.

TomBelmar:


src/clue.js
```diff
@@ -2,25 +2,147 @@
...
+},
+];
 
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(elements) {
+    
+   people = elements[Math.floor(Math.random()*elements.length)];
```
webLisbonTA, line 126: You should always initialize the variable with let or const. So, it's better pratice if you did 
`let people = elements[Math.floor(Math.random()*elements.length)];`.
However, you didn't even need to initialize any variable, as you could just 
`return elements[Math.floor(Math.random()*elements.length)];`

webLisbonTA (commented):
Good job on the lab Tomás, please check the comments I left you for a better understanding.

PR #3105, opened 2023-01-18
[LISFT123] Maria Carvalho Solved lab

miacsilva:


src/clue.js
```diff
@@ -2,25 +2,126 @@
...
+weight: 20}
+];
 
 
 // ITERATION 2
 
-function selectRandom() {}
-
-function pickMystery() {}
+function selectRandom(arr) {
```
joaopdg, line 102: There's no need for the two `conditions` here, because if the `array` is empty it will already return `undefined`, and if the `array` only has one `element` the `Math.floor(Math.random())` will return that `element`:

```
function selectRandom(arr) {
    return arr[Math.floor(Math.random() * arr.length)];
}
```

src/clue.js
```diff
@@ -2,25 +2,126 @@
...
+    if (arr.length===1){
+        return arr[0] }
+    else{
+        return arr[Math.floor(Math.random()*arr.length)]
+    }
+}
 
 
-// ITERATION 3
+function pickMystery() {
```
joaopdg, line 112: Here you can call the `selectRandom()` function straight away as the `value` of the `keys`:

```
  function pickMystery() {
    return {
    };
  }
```

joaopdg (commented):
Great job on the lab Maria!

PR #3107, opened 2023-01-18
[LISFT0123] Francisco Pereira

FranciscoManuelPereira:
-Solved Lab

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,25 +2,129 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: "Jacob",
+    lastName: "Green",
+    occupation: "Entrepreneur",
+    age: 45,
+    description: "He has a lot of connections",
+    image:
+      "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
+    color: "green",
+  },
+  {
+    firstName: "Doctor",
+    lastName: "Orchid",
+    occupation: "Scientist",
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,25 +2,129 @@
...
+  { name: "axe", weight: 15 },
+  { name: "bat", weight: 13 },
+  { name: "trophy", weight: 25 },
+  { name: "pistol", weight: 20 },
+];
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
```
joaopdg, line 103: There's no need for the two `conditions` here, because if the `array` is empty it will already return `undefined`, and if the `array` only has one `element` the `Math.floor(Math.random())` will return that `element`:

```
  function selectRandom(arr) {
    const randomItem = [Math.floor(Math.random() * arr.length)];
    return arr[randomItem];
  }
```

src/clue.js
```diff
@@ -2,25 +2,129 @@
...
+  if (arr.length === 1) {
+    return arr[0];
+  }
 
-function pickMystery() {}
+  const randomItem = [Math.floor(Math.random() * arr.length)];
+  return arr[randomItem];
+}
 
+function pickMystery() {
```
joaopdg, line 115: Here you can call the `selectRandom()` function straight away as the `value` of the `keys`:

```
  function pickMystery() {
    return {
    };
  }
```

joaopdg (commented):
Great job on the lab Francisco!

PR #3108, opened 2023-01-18
[LISFT 0123] João Carneiro

joao621119:


src/clue.js
```diff
@@ -2,25 +2,184 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
```
joaopdg, line 5: In this case there's no need to assign a name to the objects ( "MrGreen = { }" , "drOrchid = { }" , ... ):

```
const suspectsArray = [
  {
    firstName: "Jacob",
    lastName: "Green",

    (...)

  },
  {
    firstName: "Doctor",
    lastName: "Orchid",

    (...)

  },

    (...)
```

src/clue.js
```diff
@@ -2,25 +2,184 @@
...
+        name: "pistol",
+        weight: 20
+    }
+];
 
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
```
joaopdg, line 156: There's no need for the two `conditions` here, because if the `array` is empty it will already return `undefined`, and if the `array` only has one `element` the `Math.floor(Math.random())` will return that `element`:

```
  function selectRandom(arr) {
  let randomEl = Math.floor(Math.random() * arr.length);
  return arr[randomEl];
  }
```

src/clue.js
```diff
@@ -2,25 +2,184 @@
...
+    if(!arr.length) return undefined
+    if(arr.length === 1) return arr[0]
+    let randomEl = Math.floor(Math.random() * arr.length)
+    return arr[randomEl]
+}
 
-function pickMystery() {}
+function pickMystery() {
```
joaopdg, line 163: In the previous iteration you created a `function` that already returns you a random `element` of an `array` so you can use it inside this new `function` and in this way you keep it `DRY`:

```
function pickMystery() {
  return {
  };
}
```

joaopdg (commented):
Great job on the lab João!

PR #3109, opened 2023-01-18
[LISFT0123] Vanessa Vieira

vanessav22:


src/clue.js
```diff
@@ -2,25 +2,125 @@
...
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
+    
+        if (arr.length === 0)
+           return undefined
+        
+        let pickelement = Math.floor(Math.random() * arr.length)
```
webLisbonTA, line 101: You didn't even need to initialize a variable, as you could just
`return arr[Math.floor(Math.random()*arr.length)];`

webLisbonTA (commented):
Great job on the lab Vanessa.

PR #3115, opened 2023-01-18
[LISFT0123]Farid

Farid-97:


src/clue.js
```diff
@@ -2,25 +2,155 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+    MrGreen={
```
joaopdg, line 6: In this case there's no need to assign a name to the objects ( "MrGreen = { }" , "drOrchid = { }" , ... ):

```
const suspectsArray = [
  {
    firstName: "Jacob",
    lastName: "Green",

    (...)

  },
  {
    firstName: "Doctor",
    lastName: "Orchid",

    (...)

  },

    (...)
```

src/clue.js
```diff
@@ -2,25 +2,155 @@
...
+{name: "Guest House" 
+},
+{name: "Patio"
+}
+];
 
 // Weapons Array
 
-const weaponsArray = [];
+const weaponsArray = [
```
joaopdg, line 102: Here you don't need the " --- " in the name of the weapon, it's just the name:

```
const weaponsArray = [
  { name: "rope", weight: 10 },
  { name: "knife", weight: 8 },

  (...)

];
```

joaopdg (commented):
Great job on the lab Farid!

PR #3116, opened 2023-01-18
final version

evelinecoen:


src/clue.js
```diff
@@ -2,25 +2,148 @@
...
 
 // ITERATION 2
 
-function selectRandom() {}
 
-function pickMystery() {}
 
+function selectRandom(arr) {
+
+    if (arr.length === 0)
```
lzaquine, line 114: Since we know this array will not return undefined, and later you return `arr[pickelElement]`, you don't need these if statements.

lzaquine (commented):
Great job on the lab Eveline!

PR #3117, opened 2023-01-18
[LISFT0123] Rafaela Umbelino

RafaelaUmbelino:


src/clue.js
```diff
@@ -2,25 +2,170 @@
...
+
+
+function pickMystery() {
 
-function pickMystery() {}
+const mistery = {
+    suspect: "",
+    weapon: "",
+    room: ""
+};
```
webLisbonTA, line 145: Here you didn't need to declare the variable, you should've just
```
return {
    };
```

src/clue.js
```diff
@@ -2,25 +2,170 @@
...
+const selectMistery = pickMystery();
+
+function revealMystery(selectMistery) {
+
+
+    return `${selectMistery.suspect.firstName} ${selectMistery.suspect.lastName} killed Mr. Boddy using the ${selectMistery.weapon.name} in the ${selectMistery.room.name}!`;
+
+}
+
+console.log(revealMystery(selectMistery));
```
webLisbonTA, line 170: Everything is correct here, but you didn't need to create the variable. When doing this console.log, you could just pass the pickMystery function as an argument `console.log(revealMystery(pickMystery()));`

webLisbonTA (commented):
Great job on the lab Rafaela.

PR #3122, opened 2023-01-18
[LISFT0123] Gabriel

gbrlcsgn:


src/clue.js
```diff
@@ -2,25 +2,136 @@
...
+    {name: "bat", weight: 13},
+    {name: "trophy", weight: 25},
+    {name: "pistol", weight: 20}
+];
 
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
```
joaopdg, line 103: There's no need for the two `conditions` here, because if the `array` is empty it will already return `undefined`, and if the `array` only has one `element` the `Math.floor(Math.random())` will return that `element`:

```
  function selectRandom(arr) {
  let pickelement = Math.floor(Math.random() * arr.length);
  return arr[pickelement];
  }
```

src/clue.js
```diff
@@ -2,25 +2,136 @@
...
+    
 
-function pickMystery() {}
+    let pickelement = Math.floor(Math.random() * arr.length)
+       return arr[pickelement]
+    
+}
 
 
+function pickMystery() {
```
joaopdg, line 119: Here you can call the `selectRandom()` function straight away as the `value` of the `keys`:

```
  function pickMystery() {
    return {
    };
  }
```

joaopdg (commented):
Great job on the lab Gabriel!

PR #3233, opened 2023-04-12
[LIS-WDFT-APR23] Pedro Nogueira

Pedro-No:


src/clue.js
```diff
@@ -2,25 +2,141 @@
...
 
-function selectRandom() {}
+function selectRandom(randomize) {
+    let randomnumber = Math.floor(Math.random() * (randomize.length));
+    if (randomize.length === 0 || randomize.length === undefined) {
+        return undefined;
+    }else if(randomize.length === 1){
+        return randomize[0];
+    }else {
+        return randomize[randomnumber];
```
joaopdg, line 110: This is correct but actually there is no need for the conditions here, you can just do:

```
function selectRandom(randomize) {
           let randomnumber = Math.floor(Math.random() * (randomize.length));
           return randomize[randomnumber];
}
```

Because if the array is empty `randomIndex` will be equal to `0` and since there is no index 0 `[0]` it will already return undefined. And if we only have 1 element inside the array `randomIndex` will be `0` again but this time we have the index 0 `[0]` so we will return it.

joaopdg (commented):
Great job on the lab Pedro!

PR #3237, opened 2023-04-12
[LIS WDFT APR23 ] Bernardo Abreu Lab 3

BernardoGuedesAbreu:


src/clue.js
```diff
@@ -2,25 +2,131 @@
...
+        {name: "pistol" , weight: 20},  
+
+];
 
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
+   const values = Object.values(arr);
```
brunorocha20, line 100: This step here (`Object.values(arr);`) is unnecessary, you can just use the argument itself.
```
let values = Math.floor(Math.random() * arr.length);
return arr[values]
```

brunorocha20 (commented):
Great job on the lab Bernardo.

PR #3239, opened 2023-04-12
[LIS-WD-FT-APR-2023] Margarida Pereira

mapmargaridapereira:


src/clue.js
```diff
@@ -2,25 +2,117 @@
...
 
-function selectRandom() {}
-
-function pickMystery() {}
+function selectRandom(array) {
+  if (array.length === 0) {
+    return undefined;
+  } else {
+    let randomNum = Math.floor(Math.random() * array.length);
+    return array[randomNum];
```
joaopdg, line 106: This is correct but actually there is no need for the conditions here, you can just do:

```
function selectRandom(array) {
    let randomNum = Math.floor(Math.random() * array.length);
    return array[randomNum];
}
```

joaopdg (commented):
Great job on the lab Margarida!

PR #3240, opened 2023-04-12
[LIS-WD-FT-APR-2023] Renato Pais

Renatopais:


src/clue.js
```diff
@@ -2,25 +2,124 @@
...
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(cards) {
+ if (cards.length === 0) {
+    return undefined
+ } else {
+    let randomNum = Math.floor(Math.random() * cards.length)
+    return cards[randomNum]
```
joaopdg, line 109: This is correct but actually there is no need for the conditions here, you can just do:

```
function selectRandom(cards) {
    let randomNum = Math.floor(Math.random() * cards.length)
    return cards[randomNum]
}
```

joaopdg (commented):
Great job on the lab Renato!

PR #3245, opened 2023-04-12
[Lis-wdft-2023] - Lucas Telles

lltelles:


src/clue.js
```diff
@@ -2,25 +2,78 @@
...
 
-function selectRandom() {}
+function selectRandom(arr) {
+    if (arr.length === 0){
+        return undefined;
+    } else if (arr.length === 1){
+        return arr[0];
+    } else {
+        const randomIndex = Math.floor(Math.random() * arr.length);
+        return arr[randomIndex];
```
joaopdg, line 59: This is correct but actually there is no need for the conditions here, you can just do:

```
function selectRandom(arr) {
           const randomIndex = Math.floor(Math.random() * arr.length);
           return arr[randomIndex];
}
```

Because if the array is empty `randomIndex` will be equal to `0` and since there is no index 0 `[0]` it will already return undefined. And if we only have 1 element inside the array `randomIndex` will be `0` again but this time we have the index 0 `[0]` so we will return it.

joaopdg (commented):
Great job on the lab Lucas!

PR #3249, opened 2023-04-12
[LIS-FT-APR23] ROSHAN PRADHAN

roseanprime:


src/clue.js
```diff
@@ -1,26 +1,139 @@
...
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(array) {
+    if (array.length === 0){
+        return undefined;
+    }
+        else {
+           let randomNum = Math.floor(Math.random() * array.length);
+           return array[randomNum];
```
joaopdg, line 118: This is correct but actually there is no need for the conditions here, you can just do:

```
function selectRandom(array) {
           let randomNum = Math.floor(Math.random() * array.length);
           return array[randomNum];
}
```

joaopdg (commented):
Great job on the lab Roshan!

PR #3254, opened 2023-04-12
Lis-WDFT-APR-2023-Nuno-Durao

NunoDurao:


src/clue.js
```diff
@@ -2,25 +2,119 @@
...
 
-function pickMystery() {}
+
+function selectRandom(randomArray) {
+    if(randomArray.length === 0) {
+        return undefined
+    } 
+    const randomIndex = Math.floor(Math.random()*randomArray.length);
+    const ramdomElement = randomArray[randomIndex];
+    return ramdomElement;
```
joaopdg, line 97: This is correct but actually there is no need for the conditions here, you can just do:

```
function selectRandom(randomArray) {
  const randomIndex = Math.floor(Math.random() * randomArray.length);
  const ramdomElement = randomArray[randomIndex];
  return ramdomElement;
}
```

Because if the array is empty `randomIndex` will be equal to `0` and since there is no index 0 `[0]` it will already return undefined.

joaopdg (commented):
Great job on the lab Nuno!

PR #3257, opened 2023-04-13
wdft April 23 Gustavo Couto

Guss-Gustavo-Couto:
estava doente não consegui sequer começar o de Html/Css farei hoje

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,25 +2,54 @@
 
 // Suspects Array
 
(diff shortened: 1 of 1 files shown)


brunorocha20 (commented):
Good job on the lab Gustavo.
You had that `DATA.md` file with the correct information for the 3 arrays. 
You should've used that instead of creating it yourself.

PR #3284, opened 2023-04-18
[PTMEX042023]-Walter-Bernal

WaltBernalM:
done

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -203,5 +203,6 @@ function pickMystery() {
 
 // ITERATION 3
 
-function revealMystery() {}
-
+function revealMystery(object) {
+  return `${object.suspect.name} killed Mr. Boddy using the ${object.weapon.name} in the ${object.room.name} `
+}
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,7 +2,8 @@
 
 // Suspects Array
 let mrGreen = {
-  name: "Jacob Green",
+  firstName: "Jacob ",
+  lastName: "Green",
   occupation: "Entrepreneur",
   age: 45,
   description: "He has a lot of connections",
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -1,26 +1,216 @@
...
+  mrsPeacock,
+  mrMustard
+]
 
-const suspectsArray = [];
 
 // Rooms Array
 
-const roomsArray = [];
+let diningRoom = {
```
ta-web-mexpt2:
```suggestion
 {name: "Dining Room"},
 {name: "Conservatory"},
    {name: "Kitchen"},
```
inside of rooms array and you are not to set the variables for all the rooms the same with the others

src/clue.js
```diff
@@ -1,26 +1,216 @@
...
+  observatory,
+  theater,
+  guestHouse,
+  patio
+]
 
 // Weapons Array
 
-const weaponsArray = [];
+let rope = {
```
ta-web-mexpt2: same here

```suggestion
 {name: "rope",
 weight: 10,}
```

src/clue.js
```diff
@@ -1,26 +1,216 @@
...
+    room: selectRandom(roomsArray),
+    weapon: selectRandom(weaponsArray)
+  }
+}
 
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(object) {
```
ta-web-mexpt2: better leave the @param with sintactic sugar 

```suggestion
function revealMystery(mistery) {
```

ta-web-mexpt2 (changes requested):

WaltBernalM (commented):
I've refactored all the code to follow your feedback

PR #3285, opened 2023-04-19
[PTMEX042023]-Andres-Franco

Andresfrla:


src/clue.js
```diff
@@ -2,25 +2,147 @@
...
+    return objectCard;
+  }
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(reveal) {
+    const {suspect, weapon, room} = reveal;
+    
+    return `${reveal.suspect.firstName} ${reveal.suspect.lastName} killed Mr. Boddy using the ${reveal.weapon.name} in the ${reveal.room.name}!`
```
ta-web-mexpt2, line 146:
```suggestion
    return `${suspect.firstName} ${suspect.lastName} killed Mr. Boddy using the ${weapon.name} in the ${room.name}!`
```
since you already destructure from reveal object with 
```
 const {suspect, weapon, room} = reveal;
```

PR #3338, opened 2023-07-12
[LISFTJULY23] FILIPA FLORA LAB 3.1

filipaflora:


jacobviv:
Really good job with the lab Filipa!

src/clue.js
```diff
@@ -4,23 +4,222 @@
...
+      image: 'http://www.radiotimes.com/uploads/images/Original/111967.jpg',
+      color: 'white'
+    },
+    weapon: { name: 'dumbbell', weight: 30 },
+    room: { name: 'Spa' }
+  }
+
+function revealMystery(obj) {
+    return `${obj.suspect.firstName} ${obj.suspect.lastName} killed Mr. Boddy using the ${obj.weapon.name} in the ${obj.room.name}!`
+}
```
jacobviv, line 222: Here, you have everything correct, however, it is good practice if you destructure your `obj` argument.
By extracting the properties into individual variables (suspect, weapon, and room), you can directly reference them in your return statement without needing to use `obj.suspect`, `obj.weapon`, and `obj.room` every time. This simplifies the code and improves readability.

```
function revealMystery(obj) {
  let { suspect, weapon, room } = obj;

}
```

PR #3339, opened 2023-07-12
[LIS-FT-JULY23]Miguel Leite

AakaiLeite:


src/clue.js
```diff
@@ -4,23 +4,144 @@
...
+    mystery.room = selectRandom(roomsArray);
 
+    return mystery;
+}
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(mystery) {
+    let firstName = mystery.suspect.firstName;
```
brunorocha20, line 139: Here, you have everything correct, however, it is good practice if you destructure your `mystery` argument.
By extracting the properties into individual variables (suspect, weapon, and room), you can directly reference them in your return statement without needing to use `mystery.suspect`, `mystery.weapon`, and `mystery.room` every time. This simplifies the code and improves readability.
```
function revealMystery(mystery) {
  let { suspect, weapon, room } = mystery;

}
```

brunorocha20 (commented):
Great job on the lab Miguel.

PR #3343, opened 2023-07-12
[LIS-FT-JULY23] Bruno_Cunha

brcunha3000:


jacobviv:
Pretty good job with the lab Bruno!

src/clue.js
```diff
@@ -2,25 +2,157 @@
...
-function pickMystery() {}
+}
 
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(envelope) {
+    return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`
+}
```
jacobviv, line 157: Here, you have everything correct, however, it is good practice if you destructure your `envelope` argument.

```

}
```

PR #3346, opened 2023-07-12
[LIS-FT-JULY23] Agnes_cludo lab

AgnesMorawska:
I would like to consult regarding this later on

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,26 +1,154 @@
 // ITERATION 1
 
+let rooms = [];
+let weapons  = [];
+
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+ 
+{
+ firstName: "Jacob",
+lastName: "Green",
+occupation: "Entrepreneur",
+age: 45,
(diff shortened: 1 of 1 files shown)


jacobviv:
Good job with the lab Agnes

src/clue.js
```diff
@@ -1,26 +1,154 @@
...
+};
 
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(discover) {
+
+return `${discover.suspect.firstName} ${discover.suspect.lastName} killed Mr. Boddy using the ${discover.weapon.name} in the ${discover.room.name}!`
+
```
jacobviv, line 152: Here, you have everything correct, however, it is good practice if you destructure your `discover` argument.
By extracting the properties into individual variables (suspect, weapon, and room), you can directly reference them in your return statement without needing to use `discover.suspect`, `discover.weapon`, and `discover.room` every time. This simplifies the code and improves readability.

```
function revealMystery(discover) {
  let { suspect, weapon, room } = discover;

}
```

PR #3347, opened 2023-07-12
[LIS-FT-JULY23] Tomás Borda de Água

Vayre047:


src/clue.js
```diff
@@ -2,25 +2,192 @@
...
+    pickedStuff.weapon = selectRandom(weaponsArray);
 
+    return pickedStuff;
+}
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(envolope) {
+
```
brunorocha20, line 190: Here, you have everything correct, however, it is good practice if you destructure your `envelope` argument.
```

}
```

src/clue.js
```diff
@@ -2,25 +2,192 @@
...
+        weight: 20
+    }
+];
 
 
 // ITERATION 2
 
-function selectRandom() {}
+function selectRandom(arr) {
+    for(let i = 0; i < arr.length; i++){
```
brunorocha20, line 168: There's no need for the `for` loop here as you just need to return the `arr` with a randomized index. The way you can control that is by doing that multiplication on the `Math.floor(Math.random() * arr.length)` doing that is saying that you want a randomized value between 0 and the length of your argument, so that's just it.

brunorocha20 (commented):
Good job on the lab Tomás.

PR #3348, opened 2023-07-12
[LIS-FT-JULY23] Maureen_Treacy

maureen-treacy:


src/clue.js
```diff
@@ -2,25 +2,204 @@
...
+    // endResult.room = "Library"
+    // endResult.weapon = "Candlestick"
+    // endResult.suspect = "mrMustard"
+    // console.log(room.push)
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery (object){
+    return `${object.suspect.firstName} ${object.suspect.lastName} killed Mr. Boddy using the ${object.weapon.name} in the ${object.room.name}!`
```
brunorocha20, line 202: Here, you have everything correct, however, it is good practice if you destructure your `object` argument.
By extracting the properties into individual variables (suspect, weapon, and room), you can directly reference them in your return statement without needing to use `object.suspect`, `object.weapon`, and `object.room` every time. This simplifies the code and improves readability.
```
  let { suspect, weapon, room } = object;

}
```

brunorocha20 (commented):
Great job on the lab Maureen.

PR #3350, opened 2023-07-12
[LIS-FT-JULY23] mariana_silva

maryanasilva:


src/clue.js
```diff
@@ -2,25 +2,121 @@
...
+    };
+    return mysteryPick
+}
 
 // ITERATION 3
 
-function revealMystery() {}
-
+function revealMystery(envelope) {
+    return (`${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`);
```
brunorocha20, line 121: Here, you have everything correct, however, it is good practice if you destructure your `envelope` argument.
```

}
```

brunorocha20 (commented):
Great job on the lab Mariana.

PR #3352, opened 2023-07-12
[LIS-FT-JULY23] André_Pinto

pinto-andre:


jacobviv:
Very good work with your lab André

src/clue.js
```diff
@@ -2,25 +2,139 @@
...
+    return selected;
+}
 
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(envelope) {
+    return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`
+
```
jacobviv, line 138: Here, you have everything correct, however, it is good practice if you destructure your `envelope` argument.

```

}
```

pinto-andre: I see. After speaking with you and Bernardo I'm actually trying to make things step by step, in order to improve readability and to make more easily understandable... But I still need practice. 
I'm thinking on being active on codewars do try it.

jacobviv: As you said, it is just a matter of practice.
Try codewars, sure, but don't forget to sleep!

PR #3353, opened 2023-07-12
[LIS-FT-JULY23] Gonçalo_Matias

Gosamat:


jacobviv:
Good job with your lab Gonçalo

src/clue.js
```diff
@@ -2,25 +2,119 @@
...
+const suspectsArray = [{
+    firstName: "Jacob",
+    lastName: "Green",
+    occupation: "Entrepreneur",
+    age: 45,
+    description: "He has a lot of connections",
+    image: "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
+    color: "green"},
+    {firstName: "Doctor",
+    lastName: "Orchid",
```
jacobviv, line 14: For the shake of readability try this approach:

```
const suspectsArray = [
    {
        firstName: "Jacob",
        lastName: "Green",
        occupation: "Entrepreneur",
        age: 45,
        description: "He has a lot of connections",
        image: "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
        color: "green"
    },
    {
        firstName: "Doctor",
        lastName: "Orchid",
```

src/clue.js
```diff
@@ -2,25 +2,119 @@
...
+return result;
+}
 
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(envelope) {
+    return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`
+
```
jacobviv, line 118: Here, you have everything correct, however, it is good practice if you destructure your `envelope` argument.

```

}
```

src/clue.js
Gosamat: going through these now, thanks for the feedback. I've noticed that I struggle a bit with readability at this stage so your approach really helps.

src/clue.js
Gosamat: Ah i see, so basically making a variable inside the function that helps clarify at a glance what sort of property is meant to be used with it.

I did wonder how would a different dev know at a glance what sort of properties I wanted to be used with my function. Didn't even consider the obvious of just describing within it.

Thanks so much for the heads-up!

src/clue.js
jacobviv: Great, keep on!

PR #3355, opened 2023-07-12
Aamir Mushtaq

Aamir269:
[should return "<FIRST NAME> <LAST NAME> killed Mr. Boddy using the <WEAPON> in the <PLACE>!"](http://127.0.0.1:5501/SpecRunner.html?spec=Function%20%22revealMystery%22%20-%20Reveal%20the%20mystery%20should%20return%20%22%3CFIRST%20NAME%3E%20%3CLAST%20NAME%3E%20killed%20Mr.%20Boddy%20using%20the%20%3CWEAPON%3E%20in%20the%20%3CPLACE%3E!%22)

tried hard but could not resolve 
kindly explain

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,25 +2,163 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
 
-// Rooms Array
+   mrGreen = {
+        
+        firstName: "Jacob",
+        lastName: "Green",
+        occupation: "Entrepreneur",
+        age: 45,
+        description: "He has a lot of connections",
+        image: "https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg",
+        color: "green"
+    },
+    drOrchid = {
+        
+firstName: "Doctor",
+lastName: "Orchid",
+occupation: "Scientist",
+age: 26,
+description: "PhD in plant toxicology. Adopted daughter of Mr. Boddy",
+image: "http://www.radiotimes.com/uploads/images/Original/111967.jpg",
+color: "white"
+    },
+    profPlum = {
+    
+firstName: "Victor",
+lastName: "Plum",
+occupation: "Designer",
+age: 22,
+description: "Billionaire video game designer",
+image: "https://66.media.tumblr.com/ee7155882178f73b3781603f0908617c/tumblr_phhxc7EhPJ1w5fh03_540.jpg",
+color: "purple"
+    },
+    missScarlet = {
+    
+firstName: "Kasandra",
+lastName: "Scarlet",
+occupation: "Actor",
+age: 31,
+description: "She is an A-list movie star with a dark past",
+image: "https://www.radiotimes.com/uploads/images/Original/111967.jpg",
+color: "red"
+  },
 
-const roomsArray = [];
+  mrsPeacock = {
+firstName: "Eleanor",
+lastName: "Peacock",
+occupation: "Socialité",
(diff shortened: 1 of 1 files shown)


jacobviv:
Hi Aamir, you should try hard also with the indentation of your code, I'll leave you some examples.

src/clue.js
```diff
@@ -2,25 +2,163 @@
...
+firstName: "Doctor",
+lastName: "Orchid",
+occupation: "Scientist",
+age: 26,
+description: "PhD in plant toxicology. Adopted daughter of Mr. Boddy",
+image: "http://www.radiotimes.com/uploads/images/Original/111967.jpg",
+color: "white"
+    },
+    profPlum = {
+
```
jacobviv, line 28: Try using the tab correctly and also the blank lines:

```
const suspectsArray = [];

const mrGreen = {
  firstName: 'Jacob',
  lastName: 'Green',
  occupation: 'Entrepreneur',
  age: 45,
  description: 'He has a lot of connections',
  image: 'https://pbs.twimg.com/profile_images/506787499331428352/65jTv2uC.jpeg',
  color: 'green'
};

const drOrchid = {
  firstName: 'Doctor',
  lastName: 'Orchid',
  occupation: 'Scientist',
  age: 26,
  description: 'PhD in plant toxicology. Adopted daughter of Mr. Boddy',
  image: 'http://www.radiotimes.com/uploads/images/Original/111967.jpg',
  color: 'white'
};
```

src/clue.js
```diff
@@ -2,25 +2,163 @@
...
+ // Rooms Array
+
+const roomsArray = [
+     {
+    name: "Dining Room"},{
+    
+    name: "Conservatory"},{
+    name: "Kitchen"},{
+    name: "Study"},{
```
jacobviv, line 78: This is also correct but should look like this:

```
const roomsArray = [
     { name: "Dining Room" },
     { name: "Conservatory" },
     { name: "Kitchen" },
```

src/clue.js
```diff
@@ -2,25 +2,163 @@
...
 // Weapons Array
 
-const weaponsArray = [];
+const weaponsArray = [
+    {name: "rope",
+    weight: 10},
+    {
+name: "knife",
+weight: 8},
+{
```
jacobviv, line 100: The same in here:

```
const weaponsArray = [
    { name: "rope", weight: 10 },
    { name: "knife", weight: 8 },
```

src/clue.js
```diff
@@ -2,25 +2,163 @@
...
+console.log(pickMystery())
+  
+
+
+function revealMystery(object) {
+    const firstName = object.suspect.Kasandra;
+    const lastName = object.suspect.Scarlet;
+    const weaponUsed = object.weapon.pistol;
+    const place = object.room.GuestHouse;
+
```
jacobviv, line 150: The object `suspect` doesn't have a Kasandra or Scarlet properties, but a `firstName` one and a `lastName` too
Object `weapon` does have a property called `name`, and the same goes for object `room`

So, in this case and in order to work your code should look like:

```
  const firstName = object.suspect.firstName;
  const lastName = object.suspect.lastName;
  const weaponUsed = object.weapon.name;
  const place = object.room.name;

  return `${firstName} ${lastName} killed Mr. Boddy using the ${weaponUsed} in the ${place}!`;
}
```

The best way though should be like this:

```

}
```

It is good practice if you destructure your `envelope` argument.

PR #3356, opened 2023-07-13
[LISFT072023] Ana Paulina

anapaulina28:


jacobviv:
Good job with your clue lab Ana Paulina!

src/clue.js
```diff
@@ -1,26 +1,121 @@
...
 
 // Weapons Array
 
-const weaponsArray = [];
+const weaponsArray = [    
+{name: "rope", weight: 10},
+{name: "knife", weight: 8},
+{name: "candlestick", weight: 2},
+{name: "dumbbell", weight: 30 },
+{name: "poison", weight: 2},
```
jacobviv, line 90: Just don't forget to use the tab

```
const weaponsArray = [    
    {name: "rope", weight: 10},
    {name: "knife", weight: 8},
```

src/clue.js
```diff
@@ -1,26 +1,121 @@
...
+    }
+  return mystery
+}
 
-function revealMystery() {}
+// ITERATION 3
 
+function revealMystery (object){
+    return `${object.suspect.firstName} ${object.suspect.lastName} killed Mr. Boddy using the ${object.weapon.name} in the ${object.room.name}!`
+}
```
jacobviv, line 121: Here, you have everything correct, however, it is good practice if you destructure your `object` argument.
By extracting the properties into individual variables (suspect, weapon, and room), you can directly reference them in your return statement without needing to use `object.suspect`, `object.weapon`, and `object.room` every time. This simplifies the code and improves readability.

```
  let { suspect, weapon, room } = object;

}
```

PR #3460, opened 2023-11-28
[WD-PT] Nadine Rangosch

nenuphars:
For the math.random I had to look at the resources a bit to see how to make it an integer, but I think it was much easier than the last lab. Let me know if there are some things I can improve.

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,25 +2,121 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: "Jacob",
+    lastName: "Green",
+    occupation: "Entrepreneur",
+    age: 45,
+    description: "He has a lot of connections",
+    image:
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,25 +2,121 @@
...
+    }
+    return mystery
+}
 
 // ITERATION 3
 
-function revealMystery() {}
-
+function revealMystery(envelope) {
+    return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`
```
iPoolito, line 121: In this part we can create variables in order to make it easier to read will be somehitng like
```suggestion
const firstName=envelope.suspect.firstName
const lastName=envelope.suspect.lastName
const weapon = envelope.weapon.name
const room = envelope.room.name

  ${firstName} ${.lastName} killed Mr. Boddy using the ${weapon} in the ${room}!!`
```

iPoolito:
The way you get the random its correct, Nice Work 🔥

PR #3461, opened 2023-11-28
[WD-PT] Christian Doeller

cDoeller:


src/clue.js
```diff
@@ -2,25 +2,123 @@
...
+  newMystery.room = selectRandom(roomsArray);
+  return newMystery;
+}
 
 // ITERATION 3
 
-function revealMystery() {}
-
+function revealMystery(envelope) {
+    return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`
```
iPoolito, line 122: In this part we can create variables in order to make it easier to read will be somehitng like
```suggestion
const firstName=envelope.suspect.firstName
const lastName=envelope.suspect.lastName
const weapon = envelope.weapon.name
const room = envelope.room.name

  ${firstName} ${.lastName} killed Mr. Boddy using the ${weapon} in the ${room}!!`
```

iPoolito:
Nice job 🔥

cDoeller:
Thx for the feedback !! :)
____

Christian Doeller
Karl-Ferlemann-Str. 56
04177 Leipzig

***@***.***
www.christiandoeller.de

PR #3462, opened 2023-11-28
[WD-PT]  Juan Morales

Juanmormar:


src/clue.js
```diff
@@ -2,25 +2,150 @@
...
 
-function pickMystery() {}
 
 
 // ITERATION 3
 
-function revealMystery() {}
 
+function revealMystery(revealArray) {     
+    return `${revealArray.suspect.firstName} ${revealArray.suspect.lastName} killed Mr. Boddy using the ${revealArray.weapon.name} in the ${revealArray.room.name}!`
```
iPoolito, line 150: In this part we can create variables in order to make it easier to read will be somehitng like
```suggestion
const firstName=envelope.suspect.firstName
const lastName=envelope.suspect.lastName
const weapon = envelope.weapon.name
const room = envelope.room.name

  ${firstName} ${.lastName} killed Mr. Boddy using the ${weapon} in the ${room}!!`
```

iPoolito:
Looks Good, Nice job 🚀

PR #3463, opened 2023-11-29
[WD-PT] Eduardo Botelho

UPSxACE:
Note: I created an utility function "getRandomInt()" at the top of clue.js file

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -1,26 +1,119 @@
+/** Get random integer
+ *
+ * Range: [0-max[ */
+function getRandomInt(max) {
+  return Math.floor(Math.random() * max);
+}
+
 // ITERATION 1
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: "Jacob",
+    lastName: "Green",
+    occupation: "Entrepreneur",
+    age: 45,
+    description: "He has a lot of connections",
+    image:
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -1,26 +1,126 @@
+/** Get random integer
+ *
+ * Range: [0-max[ */
+function getRandomInt(max) {
```
iPoolito, line 4: Nice Job 🚀

src/clue.js
```diff
@@ -1,26 +1,126 @@
...
+  const room = selectRandom(roomsArray);
+  return { suspect, weapon, room };
+}
 
 // ITERATION 3
 
-function revealMystery() {}
-
+function revealMystery(envelope) {
+  return `${envelope.suspect.firstName} ${envelope.suspect.lastName} killed Mr. Boddy using the ${envelope.weapon.name} in the ${envelope.room.name}!`;
```
iPoolito, line 125: In this part we can create variables in order to make it easier to read will be somehitng like
```suggestion
const firstName=envelope.suspect.firstName
const lastName=envelope.suspect.lastName
const weapon = envelope.weapon.name
const room = envelope.room.name

  ${firstName} ${.lastName} killed Mr. Boddy using the ${weapon} in the ${room}!!`
```

iPoolito:
Looks good with that function 🚀

src/clue.js
UPSxACE: I understand!! 😮
Will start maxing out readability!

PR #3464, opened 2023-11-29
Lab done

Natpinper:
No major issues

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,25 +2,156 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+    {
+        firstName: "Jacob",
+        lastName: "Green",
+        occupation: "Entrepreneur",
+        age: 45,
+        description: "He has a lot of connections",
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,25 +2,156 @@
...
+    
+} return mystery
+}
 
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(mystery) {
+    return `${mystery.suspect.firstName} ${mystery.suspect.lastName} killed Mr. Boddy using the ${mystery.weapon.name} in the ${mystery.room.name}!`
```
iPoolito, line 152: In this part we can create variables in order to make it easier to read will be somehitng like
```suggestion
const firstName=envelope.suspect.firstName
const lastName=envelope.suspect.lastName
const weapon = envelope.weapon.name
const room = envelope.room.name

  ${firstName} ${.lastName} killed Mr. Boddy using the ${weapon} in the ${room}!!`
```

iPoolito:
Nice job 🔥

PR #3466, opened 2023-11-29
Finish lab 

eljorgecenteno:
The last itineration was a bit tricky :)

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,25 +2,122 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+    {
+        firstName: "Jacob",
+        lastName: "Green",
+        occupation: "Entrepreneur",
+        age: 45,
+        description: "He has a lot of connections",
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,25 +2,122 @@
...
+ return randomCards
+
+}
 
 
 // ITERATION 3
 
-function revealMystery() {}
+function revealMystery(result) {     
+return `${result.suspect.firstName} ${result.suspect.lastName} killed Mr. Boddy using the ${result.weapon.name} in the ${result.room.name}!`
```
iPoolito, line 121: In this part we can create variables in order to make it easier to read will be somehitng like
```suggestion
const firstName=envelope.suspect.firstName
const lastName=envelope.suspect.lastName
const weapon = envelope.weapon.name
const room = envelope.room.name

  ${firstName} ${.lastName} killed Mr. Boddy using the ${weapon} in the ${room}!!`
```

iPoolito:
Nice job 🔥

PR #3468, opened 2023-11-29
done

mashour8:


src/clue.js
```diff
@@ -2,25 +2,123 @@
...
 
+}
 
-// ITERATION 3
 
-function revealMystery() {}
+// ITERATION 3
 
+function revealMystery(obj) {
+    return `${obj.suspect.firstName} ${obj.suspect.lastName} killed Mr. Boddy using the ${obj.weapon.name} in the ${obj.room.name}!`
```
iPoolito, line 123: In this part we can create variables in order to make it easier to read will be somehitng like
```suggestion
const firstName=envelope.suspect.firstName
const lastName=envelope.suspect.lastName
const weapon = envelope.weapon.name
const room = envelope.room.name

  ${firstName} ${.lastName} killed Mr. Boddy using the ${weapon} in the ${room}!!`
```

PR #3469, opened 2023-11-30
Solved lab - WD PT - Marta El Bay

martachat:
I had issues in Iteration2 to generate the random index. 
I tried without a loop but could never check all requirements in Jasmine, so I did the loop.

### DIFF
--- a/src/clue.js
+++ b/src/clue.js
@@ -2,25 +2,200 @@
 
 // Suspects Array
 
-const suspectsArray = [];
+const suspectsArray = [
+  {
+    firstName: 'Jacob',
+    lastName: 'Green',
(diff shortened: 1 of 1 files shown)


src/clue.js
```diff
@@ -2,25 +2,200 @@
...
+  let currentIndex = 0;
+  let selectedElement;
+
+ 
+  for (let currentIndex = 0; currentIndex <= randomIndex; currentIndex++) {
+    selectedElement = array[currentIndex];
+  }
+
+
+  return selectedElement;
```
iPoolito, line 173: Here you only need to return the random element like this 

```suggestion
 raturn array[randomIndex]
```

